PR #3, opened 2017-06-12
Initial commit DomUI wizard component

YoeriNijs:
screenshot from 2017-06-12 16-11-21

This is a first example of how to create a wizard component in DomUI. To instantiate the wizard, create a class that setups the wizard and add it later to a DomUI page.

For example, add the following to a random wizard constructor to instantiate the wizard:
```

privateWizardPopupWindow m_wizard;

public wizardGenerator() {
    m_wizard = new WizardPopupWindow("Wizard Title", setupSteps());
}

private LinkedHashMap<HashMap<String, String>, IWizardPopupStep> setupSteps() {
		LinkedHashMap<HashMap<String, String>, IWizardPopupStep> steps = new LinkedHashMap<>();

	HashMap<String, String> step1Labels = WizardPopupWindow.createWizardStepLabels("Step 1 title", null, null, null, null);
	steps.put(step1Labels, new BulkBriefTest());

	HashMap<String, String> step2Labels = WizardPopupWindow.createWizardStepLabels("Step 2 title", null, null, "Are you sure?", null);
	steps.put(step2Labels, new BulkBriefSelectCaseType(m_model.getAllCaseTypeList()));

	HashMap<String, String> step3Labels = WizardPopupWindow.createWizardStepLabels("Step 3 title", null, null, null, "Send this!");
	steps.put(step3Labels, new BulkBriefTest());

	return steps;
}
```

Lastly, create a getter for the wizard in order to add it to a DomUI page.


YoeriNijs:
Added the abstract class WizardPopupGeneratorBase. With this base, it should be easy to create a 'generator' for the WizardPopupComponent. A generator is a class that instantiates the wizard. 

Here is an example of a implementation (it could be neater for sure, this is just a demonstration ;))

```
package nl.itris.vp.pages.cad.cases.bulkbrief;

import java.util.*;

import javax.annotation.*;

import org.apache.commons.collections.map.*;

import to.etc.webapp.nls.*;

import nl.itris.vp.biz.gct.modelcache.*;
import nl.itris.vp.components.wizard.*;

/**
 * @author <a href="mailto:yoeri.nijs@itris.nl">Yoeri Nijs</a>
 * Created on 2-6-17.
 */
@DefaultNonNull
public class BulkBriefWizardGenerator extends WizardPopupGeneratorBase {

	private static final BundleRef BUNDLE = BundleRef.create(BulkBriefWizardGenerator.class, "messages");

	private WizardPopupWindow m_wizard;

	private Map<String, Object> m_stepStorage = new HashedMap();

	private static final String TITLE = BUNDLE.getString("casebulk.title");

	private GctMetaModel m_model;

	public BulkBriefWizardGenerator(@Nonnull GctMetaModel model) {
		m_model = model;
		m_wizard = new WizardPopupWindow(TITLE, setupSteps());
	}

	private Map<Map<String, String>, IWizardPopupStep> setupSteps() {
		Map<Map<String, String>, IWizardPopupStep> steps = new LinkedHashMap<>();

		Map<String, String> step1Labels = WizardPopupWindow.createWizardStepLabels(BUNDLE.getString("casebulk.step1.title"), null, null, null, null);
		steps.put(step1Labels, new BulkBriefSelectCaseType(this, m_model.getAllCaseTypeList()));

		Map<String, String> step2Labels = WizardPopupWindow.createWizardStepLabels(BUNDLE.getString("casebulk.step2.title"), null, null, "Weet je dit heel zeker?", null);
		steps.put(step2Labels, new BulkBriefSelectCaseType(this, m_model.getAllCaseTypeList()));

		Map<String, String> step3Labels = WizardPopupWindow.createWizardStepLabels(BUNDLE.getString("casebulk.step3.title"), null, null, null, null);
		steps.put(step3Labels, new BulkBriefSelectCaseType(this, m_model.getAllCaseTypeList()));

		Map<String, String> step4Labels = WizardPopupWindow.createWizardStepLabels(BUNDLE.getString("casebulk.step4.title"), null, null, "blablabla", null);
		steps.put(step4Labels, new BulkBriefSelectCaseType(this, m_model.getAllCaseTypeList()));

		Map<String, String> step5Labels = WizardPopupWindow.createWizardStepLabels(BUNDLE.getString("casebulk.step5.title"), null, null, null, null);
		steps.put(step5Labels, new BulkBriefSelectCaseType(this, m_model.getAllCaseTypeList()));

		Map<String, String> step6Labels = WizardPopupWindow.createWizardStepLabels(BUNDLE.getString("casebulk.step6.title"), null, null, null, "Versturen");
		steps.put(step6Labels, new BulkBriefTest());

		return steps;
	}

	@Nullable
	@Override
	public Map<String, Object> getStepStorage() {
		return m_stepStorage;
	}

	@Override
	public void addToStepStorage(@Nonnull Object object) {
		m_stepStorage.put(getWizard().getCurrentStep().getStepName(), object);
	}

	@Override
	public void removeFromStepStorage() {
		m_stepStorage.remove(getWizard().getCurrentStep().getStepName());
	}

	@Override
	public WizardPopupWindow getWizard() {
		return m_wizard;
	}
}

```

fjalvingh:
Well, you might as well learn that I can nag a lot, so sorry about that, already ;)

Let's start with the proposed UI. You have put the labels for the different steps on top of the screen. I'm not too sure that this is a good idea, because this means that you have quite a limit: the total text that you can show there is quite limited. This means that:

* if the number of steps increase you might need to make the labels smaller
* if the labels need to be big they fall off the screen

A more stable interface would put the steps in the side:

image

You should also visually separate the previous/next buttons from any "finish" or "cancel" button.

A second thing is the code interface. You propose a class which implements setupSteps(), which returns a map of steps. I think that's not a good plan ;) The map loses order of the steps (you do not know which step comes first), and there is no reason to separate the step names from the step UI itself.

A better way would be to have the main Wizard dialog class have a method "addStep()" which accepts some Step, and handles what to do with the step inside the UI. This would also allow you to easily update the UI if (as you seem to propose) steps can be added/changed dynamically. I myself would not allow that, because the whole point of having a wizard is to make clear what steps to take, so changing them is a bad idea because you lied to the user ;)

The wizard steps themselves are a thing too. You should wonder what a step needs to be or needs to do. Some things that are related to its use are:

* A step has a UI, which is some NodeContainer
* The step should know when it's complete, so that pressing the next button goes to the next step
* It should know whether it's OK to be opened from "back" (because usually not all wizard steps can be redone)
* A step has its own state, being the data that has been entered (so far)

I think I saw you step being implemented by an interface. If it remains just that then you get a klunky code UI: the interface must then somehow also have a createUI thing, and now the interface must talk with the result of this.

A better solution would be to make a step be a NodeContainer, either by making a step require a given abstract base class (AbstractWizardStep) with abstract methods defining the requirements, or by defining the add() method for a step to use a compound type, like NodeBase & IStepInterface

to.etc.domui/src/main/java/to/etc/domui/component/wizard/WizardPopupStepNavbarFragment.java
```diff
@@ -0,0 +1,39 @@
...
+ * Created on 2-6-17.
+ */
+@DefaultNonNull
+public class WizardPopupStepNavbarFragment extends Div {
+
+	private String m_name;
+
+	private boolean m_current;
+
+	WizardPopupStepNavbarFragment(@Nonnull WizardPopupStep step, boolean isCurrentStep) {
```
fjalvingh: Being the current step should not be a property of this class, and also not be a part of the constructor, since who is current changes.
Having current in this class means that when the current thing changes you need to walk all other instances, setting current to false, before setting this to true.

The "current step" should be a property of the main wizard class, and by binding that property to this instance you can make the updating automatic.

to.etc.domui/src/main/java/to/etc/domui/component/wizard/WizardPopupStepNavbarFragment.java
```diff
@@ -0,0 +1,39 @@
...
+		m_current = isCurrentStep;
+	}
+
+	@Override
+	public void createContent() throws Exception {
+		Div stepName = new Div();
+		stepName.setCssClass("steps-list-item");
+
+		if(m_current) {
+			LinkButton lb = new LinkButton(m_name);
```
fjalvingh: Having the current step be a link button is useless: you are already there ;) It is more logical to have the "previous" steps be a link button, so that you can go back to any step with one click.

to.etc.domui/src/main/java/to/etc/domui/component/wizard/WizardPopupWindow.java
```diff
@@ -0,0 +1,193 @@
...
+import javax.annotation.*;
+import java.util.*;
+
+/**
+ * This class holds the WizardPopupWindow for creating a dynamic wizard window.
+ * @author <a href="mailto:yoeri.nijs@itris.nl">Yoeri Nijs</a>
+ * Created on 2-6-17.
+ */
+@DefaultNonNull
+public class WizardPopupWindow extends Window {
```
fjalvingh: Should probably be final.

to.etc.domui/src/main/java/to/etc/domui/component/wizard/WizardPopupWindow.java
```diff
@@ -0,0 +1,193 @@
...
+
+	private static final String STEP_LABEL_END = "end";
+
+	private final ArrayList<WizardPopupStep> m_steps = new ArrayList<>();
+
+	private WizardPopupStep m_currentStep;
+
+	private Div m_currentStepBody;
+
+	public WizardPopupStep getCurrentStep() {
```
fjalvingh: Yes, this is how "current" should be done ;)

to.etc.domui/src/main/java/to/etc/domui/component/wizard/WizardPopupWindow.java
```diff
@@ -0,0 +1,193 @@
...
+		return m_currentStep;
+	}
+
+	public void setCurrentStep(@Nonnull WizardPopupStep step) {
```
fjalvingh: Should be very much private, or should to an awful lot more.

to.etc.domui/src/main/java/to/etc/domui/component/wizard/WizardPopupWindow.java
```diff
@@ -0,0 +1,193 @@
...
+		m_currentStep = step;
+	}
+
+	public ArrayList<WizardPopupStep> getSteps() {
```
fjalvingh: Never use ArrayList as a reference type; use List.

to.etc.domui/src/main/java/to/etc/domui/component/wizard/WizardPopupWindow.java
```diff
@@ -0,0 +1,193 @@
...
+	@Override
+	public void createContent() throws Exception {
+		refreshNavbarContainer();
+		refreshBodyContainer();
+	}
+
+	/**
+	 * This refreshes the div that holds the wizard's navbar.
+	 */
+	public void refreshNavbarContainer() {
```
fjalvingh: You'd better not refresh twice... Either the name is wrong or you should ensure that only one navbar is present.

Instead of using refresh, try to use data binding so that updating is automatic?

to.etc.domui/src/main/java/to/etc/domui/component/wizard/WizardPopupWindow.java
```diff
@@ -0,0 +1,193 @@
...
+		navbar.add(stepsContainer);
+		add(navbar);
+	}
+
+	/**
+	 * This refreshes the div that holds the body of the current step.
+	 * @throws Exception
+	 */
+	public void refreshBodyContainer() throws Exception {
+		if(null == m_currentStepBody) {
```
fjalvingh: This one is correct as far as the refreshing goes.

to.etc.domui/src/main/java/to/etc/domui/component/wizard/WizardPopupWindow.java
```diff
@@ -0,0 +1,193 @@
...
+			m_currentStepBody = new Div();
+		}
+		m_currentStepBody.removeAllChildren();
+		m_currentStepBody.add(new WizardPopupStepBodyFragment(m_currentStep, getActions(), isFirstStep(), isLastStep()));
```
fjalvingh: You should not have to new anything here; the steps should already be there somehow? What happens when you go forth and back?

to.etc.domui/src/main/java/to/etc/domui/component/wizard/WizardPopupWindow.java
```diff
@@ -0,0 +1,193 @@
...
+		return phasesContainer;
+	}
+
+	/**
+	 * Create list of available wizard actions (e.g. next button, end button).
+	 * @return
+	 */
+	private Map<WizardPopupStepActions, IUIAction<Void>> getActions() {
+		Map<WizardPopupStepActions, IUIAction<Void>> actions = new HashMap<>();
+		for(WizardPopupStepActions action : WizardPopupStepActions.values()) {
```
fjalvingh: Why have an enum? Adding the actions here is way cleaner?

to.etc.domui/src/main/java/to/etc/domui/component/wizard/WizardPopupWindow.java
```diff
@@ -0,0 +1,193 @@
...
+	 * With this, one can create wizard step labels. Only the step title label is mandatory.
+	 * If other labels are not provided, the default labels will be used.
+	 * @param titleLabel
+	 * @param cancelButtonLabel
+	 * @param backButtonLabel
+	 * @param nextButtonLabel
+	 * @param endButtonLabel
+	 * @return
+	 */
+	public static Map<String, String> createWizardStepLabels(@Nonnull String titleLabel, @Nullable String cancelButtonLabel, @Nullable String backButtonLabel, @Nullable String nextButtonLabel, @Nullable String endButtonLabel) {
```
fjalvingh: This seems to repeat. You should definitely not need a map, because what labels to use is very determinate; having fields is way better.

to.etc.domui/src/main/java/to/etc/domui/component/wizard/messages.properties
```diff
@@ -0,0 +1,4 @@
+wizardstep.default.backbutton=Previous
+wizardstep.default.nextbutton=Next
+wizardstep.default.cancelbutton=Cancel
+wizardstep.default.endbutton=Done
```
fjalvingh: The usual name for this step is finish, not done, because finish implies that an action is being taken while done does not.

to.etc.domui/src/main/resources/resources/themes/css-domui/670_wizard.frag.css
```diff
@@ -0,0 +1,34 @@
+/** Generic Wizard Component **/
+div.wizard-navbar {
```
fjalvingh: In DomUI the convention for class names is: make a short name for the component (like gwiz). Then use ui-gwiz-xxxx as class names. This makes clear where labels come from and prevents name clashes with other style sheets.

to.etc.domui/src/main/resources/resources/themes/css-domui/670_wizard.frag.css
```diff
@@ -0,0 +1,34 @@
...
+    position: absolute;
+    width: 100%;
+    top: 28px;
+    left: 0;
+    display: flex;
+}
+
+div.wizard-body {
+    margin-top: 40px;
```
fjalvingh: I assume that you want to make a fixed top bar with a flex container for the step body.
To do that you make the top container position as follows:
top: 0px
bottom: 28px
left: 0px
right: 0px
position: absolute
In your code here the top bar behaves oddly: the top is 28 pixels below the dialog start, and it can grow vertically if the line becomes too long. So the div does not overlay the full parent part.

The container for the step should be:
position: absolute
top: 28px
left: 0px
right: 0px
bottom: 0px

This makes the bottom part resize fully, with the top part being exactly 28 pixels high and overlaying the full parent area.

YoeriNijs:
According to our UX-expert, it is questionable if it is handy to place the navbar on the side, because, as a consequence, the width on the right is limited. This might be inconvenient for displaying data tables, for instance. However, as our UX guy noted as well, both versions exist in the wild. Therefore, I believe this component should support both types, so DomUI users can choose which type of wizard they want. 

Just my two cents. What's your opinion on this?

fjalvingh:
A wizard should not need to show big tables in its panels, because that implies that we do way more than would be reasonable for a wizard. In addition: screens have become wide nowadays; the old 4:3 aspect ratio is nowadays usually 16:9, so the screen is wider than before.
The advantage of having the steps on the left is also that it scales and degrades easily: when a single step's text becomes too big for the area it nicely wraps to the next line; when there are too many steps for the vertical space it scroll - all without any work.
Compare that with the horizontal approach - nothing there will work reasonably there when one of the above occurs, so it always incurs quite some work to get it "right".
And finally, marking that steps are completed is easier doing it on the left because again the space taken will not cause troubles.

So from an efficiency point of view I would use the left-side approach.

That said, it would be nice of course if you can support both approaches, and it would be a great exercise to do that in the nicest way possible, with only a little bit of code that differs between both approaches ;)

PR #4, opened 2017-06-15
Fixed clear button in LookupInput

molshoop:
When you clicked the clear button of the LookupInput it disappeared after clearing the input field. Now it's still visible and disabled.


to.etc.domui/src/main/java/to/etc/domui/component2/lookupinput/LookupInputBase2.java
```diff
@@ -804,13 +804,13 @@ public void setValue(@Nullable OT v) {
 			return;
 		OT old = m_value;
 		m_value = v;
+		getClearButton().setDisplay(DisplayType.INLINE);
```
fjalvingh: It seems to me like this is not needed as we're not changing the display type anymore.

to.etc.domui/src/main/java/to/etc/domui/component2/lookupinput/LookupInputBase2.java
```diff
@@ -804,13 +804,13 @@ public void setValue(@Nullable OT v) {
...
+		setCssClass("ui-lui2-selected");
 		if(v != null) {
-			getClearButton().setDisplay(DisplayType.INLINE);
+			getClearButton().setDisabled(false);
 			clearMessage();
-			setCssClass("ui-lui2-selected");
 		} else {
-			getClearButton().setDisplay(DisplayType.NONE);
-			setCssClass("ui-lui2");
```
fjalvingh: The original code uses ui-lui2 as style for disabled and ui-lui2-selected as value for when there is a value. You removed the ui-lui2 style so once a value is set it stays ui-lui2-selected always. That seems wrong, so why was that other class remvoved?

fjalvingh:
Will fix the issues after merge.

PR #11, opened 2017-08-18
Fixed an issue that an image will be created when string is empty

YoeriNijs:
Previously, when one sets an image with an empty string, an image was still created. This caused some problems in the UI, because DomUI still generated a div with some css, that could led to some messed up things. Because of this, in my opinion, the whole div is useless when no image is set. So now it is checked whether the string is empty.


to.etc.domui/src/main/java/to/etc/domui/component/layout/TabPanelBase.java
```diff
@@ -273,19 +77,23 @@ protected void renderTabPanels(NodeContainer labelcontainer, NodeContainer conte
 			boolean isselected = getCurrentTab() == index;
 			//-- Add the body to the tab's main div, except if it is lazy.
 			NodeBase content = ti.getContent();
-			content.addCssClass("ui-tab-pg");
-			content.setClear(ClearType.BOTH);
+			if(content != null) {
+				content.addCssClass("ui-tab-pg");
+				content.setClear(ClearType.BOTH);
```
fjalvingh: clear: both should be in the style sheet, not in code.

to.etc.domui/src/main/java/to/etc/domui/component/layout/TabPanelBase.java
```diff
@@ -382,13 +183,23 @@ public void closeTab(@Nonnull final ITabHandle th) throws Exception {
...
 
 		if(isBuilt()) {
-			ti.getTab().remove();
-			ti.getSeparator().remove();
-			ti.getContent().remove();
+			Li tab = ti.getTab();
+			if(tab != null)
+				tab.remove();
+
+			Li seperator = ti.getSeparator();
```
fjalvingh: Spelling error in separator, you might want to switch on spelling check in IntelliJ; it is quite nice ;)

to.etc.domui/src/main/java/to/etc/domui/util/DomUtil.java
```diff
@@ -79,6 +79,21 @@ private DomUtil() {}
 	}
 
 	/**
+	 * Does explicit non null check for parameter passed in as nullable but expected to be non null.
```
fjalvingh: Use Objects.requireNonNull instead: this method is no longer needed.

PR #23, opened 2017-09-29
Domutil compare and post js arrays

millevlada:
Hi, added some (I hope) nice utility for simplifying comparator functions as we use a lot of that. It reduced boiler plate code. 
I also altered few existing compares to show the use case(s).

Also added some handy fix that we needed in project to support POST requests due to large amount of data, for url that also contains String array as value (repeated names of parameter with x1000 of values).
Also POST request now can use _blank as target to force new browser tab/window as destination. Btw postURL was missing in domui.js, so existing code in DomUtil was broken, now it is connected again. 

Hope you'll like that ;)


fjalvingh:
Thanks a lot ;) I fixed some formatting issues and will merge now.

PR #47, opened 2017-10-25
Initial commit for another error message div style

YoeriNijs:
The idea: wrap all error messages in another div that may contain additional styling. So, instead of providing various rows that have separate icon and text styling, style the whole div with underlying text rows. In order to be compatible with old/other DomUI styles, current styling is not changed.

One can 'activate' the additional styling by providing a fragment that contains content for the corresponding css class. So, in this preview, the fragment may look like:
```

/*** errormessages.frag.css ***/
/*** This stylesheet overrides DomUI's default styling for the ErrorMessageDiv, a helper ***/
/*** for components that handle their error display, with ViewPoint specific styling. ***/
html {
	height: 100%;
	width: 100%;
}
.ui-emd {
	border-radius: 2px;
	-webkit-border-radius: 2px;
	-moz-border-radius: 2px;
}
.ui-emd-msg {
	background-repeat: no-repeat;
	background-position: 10px;
	padding-left: 50px;
	background-size: 11px 11px;
}
.ui-emd-error {
	background-image: none;
	margin: 10px !important;
	padding-left: 50px !important;
}
.ui-emd-info {
	background-image: none;
	margin: 10px !important;
	padding-left: 50px !important;
}
.ui-emd-warning {
	background-image: none;
	margin: 10px !important;
	padding-left: 50px !important;
}
.ui-emd-brd-error {
	background-color: <%= red_bg %>;
	border: 1px solid <%= red_stroke %>;
	background-image: url(../../../../themes/vp-icons-clean/Default/32x32_error.png);
	background-repeat: no-repeat;
	background-position: 10px;
}
.ui-emd-brd-info {
	background-color: <%= navy_blue_bg %>;
	border: 1px solid <%= navy_blue_stroke %>;
	background-image: url(../../../../themes/vp-icons-clean/Default/32x32_info.png);
	background-repeat: no-repeat;
	background-position: 10px;
	margin: 10px !important;
	padding: 10px !important;
}
.ui-emd-brd-warning {
	background-color: <%= orange_bg %>;
	border: 1px solid <%= orange_stroke %>;
	background-image: url(../../../../themes/vp-icons-clean/Default/32x32_warning.png);
	background-repeat: no-repeat;
	background-position: 10px;
	margin: 10px !important;
	padding: 10px !important;
}
```
Please note that for .ui-emd-x the background-image is none. Eventually, you may have something like this:

screenshot from 2017-10-25 14-29-03

### DIFF
--- a/to.etc.domui/src/main/java/to/etc/domui/component/layout/ErrorMessageDiv.java
+++ b/to.etc.domui/src/main/java/to/etc/domui/component/layout/ErrorMessageDiv.java
@@ -24,14 +24,15 @@
  */
 package to.etc.domui.component.layout;
 
-import java.util.*;
-
-import javax.annotation.*;
-
 import to.etc.domui.dom.css.*;
 import to.etc.domui.dom.errors.*;
 import to.etc.domui.dom.html.*;
 import to.etc.domui.util.*;
+
+import javax.annotation.*;
+import java.util.*;
+
+import static to.etc.domui.dom.css.VisibilityType.*;
 
 /**
  * This is the default in-component error handling panel, for components that
@@ -65,12 +66,12 @@ public class ErrorMessageDiv extends Div implements IErrorMessageListener {
 		}
 		DomUtil.getMessageFence(parent).addErrorListener(this);
 		setCssClass("ui-emd");
-		setVisibility(VisibilityType.HIDDEN);
+		setVisibility(HIDDEN);
 	}
 
 	public ErrorMessageDiv() {
 		setCssClass("ui-emd");
-		setVisibility(VisibilityType.HIDDEN);
+		setVisibility(HIDDEN);
 	}
 
 	public void setAsErrorFence(NodeContainer parent) {
@@ -114,6 +115,7 @@ public class ErrorMessageDiv extends Div implements IErrorMessageListener {
 	}
 
 	protected NodeContainer createErrorLine(UIMessage m) {
+		addAdditionalStyling(m);
 		Div d = new Div();
 		add(d);
 		d.setCssClass("ui-emd-msg ui-emd-" + m.getType().name().toLowerCase());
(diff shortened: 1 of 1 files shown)


to.etc.domui/src/main/java/to/etc/domui/component/layout/ErrorMessageDiv.java
```diff
@@ -126,6 +128,14 @@ protected NodeContainer createErrorLine(UIMessage m) {
 		return d;
 	}
 
+	/**
+	 * Adds css selector for additional styling to div container.
+	 */
+	private void addAdditionalStyling(@Nonnull UIMessage m) {
+		if(getChildCount() >= 0 && getVisibility() == VISIBLE)
```
fjalvingh: getChildCount() >= is always true. The visibility should not matter as far as the extra class is concerned.

to.etc.domui/src/main/java/to/etc/domui/component/layout/ErrorMessageDiv.java
```diff
@@ -126,6 +128,14 @@ protected NodeContainer createErrorLine(UIMessage m) {
...
+			addCssClass("ui-emd-brd-" + m.getType().name().toLowerCase());
```
fjalvingh: Ok, so the plan is to have the whole component expose a class that represents the "error" inside it?

This code would not work if there are multiple messages with different severity, i.e. one error, one info, one warning. In this case the "last one added" wins and that severity is exposed in the class.

The proper thing to do would be to first add the message, then walk over all messages to see what the highest severity is. Then you set the class to that severity. You also need to make sure that "earlier" classes are removed. The easiest way to do that is to actually have the component have a "severity" too and set that; then you can use the "previous severity" to remove the class and the new one to add it.

PR #48, opened 2017-10-26
Select getter added to ComboComponentBase2

YoeriNijs:
In my view, there should be a getter for the select tag in ComboComponentBase2. With the getter, it is easier to style the combo, if needed. Particularly, this might be handy if one wants to set the width of the component to 100%. In order to do that, both the combo div and the select tag need a min-width of 100%. 

E.g.:
getCombo().setCssClass("a-class-with-min-width");
getCombo().getSelect().setCssClass("another-class-with-min-width");

Of course, one can style the component with a specific css class as well (div.a-class-with-min-width > .another-class-with-min-width, or something like that), but I think this may be another clean way to adjust styling in a particular situation.

### DIFF
--- a/to.etc.domui/src/main/java/to/etc/domui/component2/combo/ComboComponentBase2.java
+++ b/to.etc.domui/src/main/java/to/etc/domui/component2/combo/ComboComponentBase2.java
@@ -24,38 +24,19 @@
  */
 package to.etc.domui.component2.combo;
 
-import to.etc.domui.component.buttons.SmallImgButton;
-import to.etc.domui.component.input.AbstractDivControl;
-import to.etc.domui.component.input.CriteriaComboDataSet;
-import to.etc.domui.component.meta.ClassMetaModel;
-import to.etc.domui.component.meta.MetaManager;
-import to.etc.domui.component.meta.PropertyMetaModel;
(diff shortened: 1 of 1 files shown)


fjalvingh:
That is a mistake we have made a lot in the past 8-/ Adding that setter exposes the internals of that component- and by doing that you need to obey all of the rules that the comnponent itself uses when playing with that part. This makes the code fragile as any change to the component (for instance not always HAVING a select) will cause the code to abort.

A correct way of doing this would be:
- add some special class to the _component_ instead of to some inner part of it
- use css to address the inner select.

For complex components make sure you use proper css selectors; making selectors too wide (like "any table inside a DataTable") will have unexpected effects when components are nested. See
https://etc.to/confluence/display/DOM/DomUI+component+rules
for some more info.

YoeriNijs:
That is a valid argument. In that case, I will css instead, thanks. Good to know that documentation is evolving rapidly by the way! 

I will close this pull request.

fjalvingh:
Well, rapidly is a big word; it is amazing how much time writing it takes. But I now try to get into the habit of immediately writing down something when I build or change it, even when nothing existed before.. So at least it grows for new stuff ;)

PR #68, opened 2018-02-06
When creating a RowRenderer with a factory, the control is not editable

benshoe:
When the factory method on the RowRenderer is used, the control inside the factory method is responsible for being editable or not.


fjalvingh:
Thanks, I now finally understand the real problem - I had a completely different idea ;)

I made some small fixes in the code after yours. Thanks for the help!

benshoe:
Glad I could help!

PR #75, opened 2018-11-05
[wip] First part of SPI code: optimizing the current request handling code from the ancient mess it is.

fjalvingh:
Refactor of the current main action handler, ancient but crucial code, into something more maintainable and understandable.


fjalvingh:
Merge fixes and simplifications part from SPI branch.

PR #80, opened 2019-07-02
Add utility method for determining visibility of an item.

stefankaticec:


to.etc.domui/src/main/java/to/etc/domui/component/tbl/DataTable.java
```diff
@@ -1109,6 +1109,10 @@ public boolean isVisible(TableRowSet<T> rowSet) {
 		return m_visibleItemList.contains(rowSet);
 	}
 
+	public boolean isVisible(@NonNull T data) {
```
fjalvingh, line 1112: Old PR, but: this will not work properly in a subtle way: when called the DataTable might not have been built yet. In that case the result will be wrong. I am not too eager to have this method, so if we no longer need it we can remove the pr.

fjalvingh (changes requested):

PR #81, opened 2019-08-02
[AceEditor] Add insert add methods.

stefankaticec:


to.etc.domui/src/main/java/to/etc/domui/component/ace/AceEditor.java
```diff
@@ -761,4 +761,17 @@ public String getCssClass() {
 	@Override public void setHint(String hintText) {
 		//setTitle(hintText);
 	}
+
+	public void insertAtCursor(String txt) throws Exception {
+		var sb = new StringBuilder();
+		sb.append("var pos = ");
+		handle(sb).append("getCursorPosition();");
+		appendJavascript(handle(sb).append("session.insert(pos, \"").append(txt).append("\");").toString());
```
fjalvingh: This would die a horrible death if the string being inserted contains a "

Use StringTool.strToJavascriptString() to properly escape scary characters in the insert string.

to.etc.domui/src/main/java/to/etc/domui/component/ace/AceEditor.java
```diff
@@ -761,4 +761,17 @@ public String getCssClass() {
...
+	}
+
+	public void insertAt(String text, int row, int column) {
+		var sb = new StringBuilder();
+		sb.append("var pos = ").append("{row: ").append(row).append(", column: ").append(column).append("};");
+		appendJavascript(handle(sb).append("session.insert(pos, \"").append(text).append("\");").toString());
```
fjalvingh: Same remark.

PR #82, opened 2019-09-05
Tab panel force rebuild

stefankaticec:
I created this branch from skarp-prd, so it contains two of your commits.

### DIFF
--- a/to.etc.domui/src/main/java/to/etc/domui/component/layout/TabPanelBase.java
+++ b/to.etc.domui/src/main/java/to/etc/domui/component/layout/TabPanelBase.java
@@ -43,6 +43,8 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 
+import static java.util.Objects.requireNonNull;
+
 public class TabPanelBase extends Div {
 	@Nullable
 	private NodeContainer m_labelContainer;
@@ -282,8 +284,15 @@ public class TabPanelBase extends Div {
 			m_tablist.add(pos, ti);
 		}
 		m_tabBuilder = null;
-		if(isBuilt())
-			forceRebuild();
+		if(isBuilt()) {
+			renderLabel(pos, ti);
+			if(!ti.isLazy()) {
+				var content = ti.getContent();
+				content.setDisplay(DisplayType.NONE);
+				ti.setAdded(true);
+				requireNonNull(m_contentContainer).add(content);
+			}
+		}
 		return ti;
 	}
 
--- a/to.etc.domui/src/main/java/to/etc/domui/component/layout/TabPanelBase.java
+++ b/to.etc.domui/src/main/java/to/etc/domui/component/layout/TabPanelBase.java
@@ -276,8 +276,9 @@ public class TabPanelBase extends Div {
 				requireNonNull(m_contentContainer).add(content);
 			}
 		}
-		if(pos <= m_currentTab)
+		if(pos <= m_currentTab && m_tablist.size() != 1) { // increment if inserted before current and this isn't the only node.
 			m_currentTab++;
+		}
 		return ti;
 	}
 
--- a/to.etc.domui/src/main/java/to/etc/domui/component/layout/TabPanelBase.java
+++ b/to.etc.domui/src/main/java/to/etc/domui/component/layout/TabPanelBase.java
@@ -122,10 +122,6 @@ public class TabPanelBase extends Div {
 	protected void renderLabel(int index, TabInstance ti) {
 		NodeContainer into = Objects.requireNonNull(m_labelContainer);
 		Li li = ti.getTab();
-//		Li separator = new Li();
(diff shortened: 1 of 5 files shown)


to.etc.domui/src/main/java/to/etc/domui/component/layout/TabPanelBase.java
```diff
@@ -282,8 +284,15 @@ private TabInstance addTabInstance(TabInstance ti, int pos) {
 			m_tablist.add(pos, ti);
```
fjalvingh: Something is wrong here in the existing code.
If the "pos" is out of bounds the thing gets added as the last node. But "pos" remains "wrong". The code should become something like this I think:

    if(pos < 0 || pos >= m_tabList.size()) {
      pos = m_tabList.size();
    }
    m_tabList.add(pos, ti);

In addition, the m_currentTab field is not correctly updated: if we insert at or before m_currentTab then m_currentTab must be incremented. I.e. if currentTab is 3, and we insert at 2, then after the insert currentTab is 4.

So

    if(pos <= m_currentTab)
      m_currentTab++;

to.etc.domui/src/main/java/to/etc/domui/dom/HtmlTagRenderer.java
```diff
@@ -677,9 +677,11 @@ public void basicNodeRender(final NodeBase b, final IBrowserOutput o, boolean in
 			UIDragDropUtil.exposeDroppable(b, dh);
 		}
 
-		String s = getStyleFor(b); 								// Get/recalculate style
-		if(s.length() > 0)
-			o.attr("style", s); 								// Append style
+		String cachedStyles = b.getCachedStyle();
```
fjalvingh: Well, the logic here is very hard to follow, so this will break easily....

I think it is still wrong. When a completely new tree is rendered then cachedStyle is also null. In that case we render a style="" on every node that has no attributes.

I might have a solution to this but it's better to talk through it. Call me when you find the time (unless you can convince me that this code is correct, of course ;)

to.etc.domui/src/main/java/to/etc/domui/component/layout/TabPanelBase.java
```diff
@@ -282,8 +284,15 @@ private TabInstance addTabInstance(TabInstance ti, int pos) {
...
 		}
 		m_tabBuilder = null;
-		if(isBuilt())
-			forceRebuild();
+		if(isBuilt()) {
+			renderLabel(pos, ti);
```
fjalvingh: renderLabel() seems to render the tab label always at the end, not at the "pos" position. So I think you need a fix there too.

to.etc.domui/src/main/java/to/etc/domui/dom/html/NodeBase.java
```diff
@@ -307,6 +309,7 @@ public void internalOnClicked(@NonNull ClickInfo cli) throws Exception {
 	final protected void changed() {
 		if(m_disableChanged > 0)
 			return;
+		setRenderStyles();
```
fjalvingh: This will again always render the styles. This method should not be called here.

to.etc.domui/src/main/java/to/etc/domui/dom/html/NodeBase.java
```diff
@@ -188,6 +188,8 @@
 	 */
 	static private final byte F_NOREPLACE = 0x08;
 
+	static private final byte F_RENDERSTYLES = 0x16;
```
fjalvingh: Should be called F_STYLERENDERED, because it is telling you not to RENDER styles but _that_ a style was rendered last time around.

to.etc.domui/src/main/java/to/etc/domui/dom/HtmlTagRenderer.java
```diff
@@ -678,7 +678,7 @@ public void basicNodeRender(final NodeBase b, final IBrowserOutput o, boolean in
 		}
 
 		String s = getStyleFor(b); 								// Get/recalculate style
-		if(s.length() > 0)
+		if(s.length() > 0 || b.shouldRenderStyles())
```
fjalvingh: Here we should set/reset the renderStyles flag, more or less like this:

```
if(s.length() > 0 || b.isStyleRendered()) {
  o.attr("style", s);
  if(s.length() == 0)
    clearStyleRendered(); // if we rendered the empty style we can remove it next time
  else
    setStyleRendered(); // otherwise there is stuff there
} 
```

to.etc.domui/src/main/java/to/etc/domui/dom/html/NodeBase.java
```diff
@@ -2290,4 +2293,14 @@ public NodeBase css(@NonNull String ...classNames) {
 		}
 		return this;
 	}
+
+	public void setRenderStyles() {
```
fjalvingh: So these names should change too.

PR #93, opened 2020-03-10
Allow upload size to be configured.

stefankaticec:


to.etc.domui/src/main/java/to/etc/domui/server/DomApplication.java
```diff
@@ -2321,6 +2323,14 @@ public void setDefaultHintsOnControl(boolean defaultHintsOnControl) {
 		m_defaultHintsOnControl = defaultHintsOnControl;
 	}
 
+	public int getMaxUploadSize() {
```
fjalvingh, line 2326: These should be either synchronized or m_maxUploadSize should be volatile.. It will not cause a problem I think because this should usually only be called at init time, and that is synchronized anyway.

fjalvingh (approved):

PR #98, opened 2020-03-31
Added result set excel exporter

millevlada:


to.etc.domui/src/main/java/to/etc/domui/util/exporters/ExcelFormat.java
```diff
@@ -39,4 +48,26 @@ static public ExcelFormat	byExtension(String ext) {
 		}
 		return null;
 	}
+
+	public static Workbook getWorkbook(ExcelFormat excelFileType) {
```
millevlada: can be changed to methods... I'll fix it.

millevlada: done

to.etc.domui/src/main/java/to/etc/domui/util/exporters/ResultSetExporter.java
```diff
@@ -0,0 +1,110 @@
...
+import java.sql.ResultSet;
+import java.sql.ResultSetMetaData;
+import java.sql.SQLException;
+import java.sql.Types;
+
+/**
+ * Basic exporter that dumps result set into workbook, as new sheet with specified name.
+ */
+@NonNullByDefault
+public class ResultSetExporter {
```
fjalvingh: It only exports to Excel, so let's call it ResultSetExcelExporter.

to.etc.domui/src/main/java/to/etc/domui/util/exporters/ResultSetExporter.java
```diff
@@ -0,0 +1,110 @@
...
+
+	private final ExcelFormat m_format;
+
+	public ResultSetExporter(ExcelFormat format) {
+		m_format = format;
+	}
+
+	public void writeExcel(Workbook workbook, ResultSet rs, int rowCount, String sheetName) {
```
fjalvingh: We should not require the #of rows, because that always means doing a possibly expensive extra query. It is easy enough to just abort if you find that you are exceeding the row count while iterating over the dataset.

fjalvingh (changes requested):

to.etc.domui/src/main/java/to/etc/domui/util/exporters/ResultSetExporter.java
millevlada: yeah, that is true... I had it for free so I used it, but you are right - it would not always be there _for free_. I'll remove that.

millevlada: ok,should be fixed now

fjalvingh (approved):

PR #99, opened 2020-04-07
Added fix for calculating pmm path and unit tests that show issues wh…

millevlada:
…en we do exists on path that has list property as one of parents in path

### DIFF
--- a/to.etc.domui.demo/src/test/java/to/etc/domui/test/db/TestDbQCriteria.java
+++ b/to.etc.domui.demo/src/test/java/to/etc/domui/test/db/TestDbQCriteria.java
@@ -6,11 +6,17 @@ import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import to.etc.domui.derbydata.db.Album;
+import to.etc.domui.derbydata.db.Album_;
 import to.etc.domui.derbydata.db.Artist;
+import to.etc.domui.derbydata.db.Artist_;
 import to.etc.domui.derbydata.db.Customer;
 import to.etc.domui.derbydata.db.Employee;
+import to.etc.domui.derbydata.db.Genre;
+import to.etc.domui.derbydata.db.Genre_;
 import to.etc.domui.derbydata.db.Invoice;
 import to.etc.domui.derbydata.db.InvoiceLine;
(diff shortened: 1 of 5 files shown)


to.etc.domui/src/main/java/to/etc/domui/component/meta/init/MetaInitializer.java
```diff
@@ -236,10 +238,24 @@ else if(score > bestscore) {
 			ix = pos + 1;
 
 			PropertyMetaModel< ? > pmm = ccmm.findSimpleProperty(sub); // Find base property,
-			if(pmm == null)
-				throw new IllegalStateException("Invalid property path '" + name + "' on " + cmm + ": property '" + sub + "' on classMetaModel=" + ccmm + " does not exist");
-			acl.add(pmm); 										// Next access path,
-			ccmm = pmm.getValueModel(); 						// was: MetaManager.findClassMeta(pmm.getActualType());
+			if(pmm == null) {
```
fjalvingh: This we cannot do here, because it invalidly maps nonexisting names onto a List and that is only valid in the context of QCriteria's "exist" subquery but nowhere else.
You need something like this, but it needs to be inside the QCriteria code, not here.

to.etc.domui.demo/src/test/java/to/etc/domui/test/db/TestDbQCriteria.java
```diff
@@ -412,5 +417,138 @@ public void testInQuery1() throws Exception {
 		Assert.assertEquals(3, ires.size());
 	}
 
+	@Test
```
fjalvingh, line 420: Thanks for the extra tests!

fjalvingh (changes requested):
Need to move the change to normal properties inside the qcriteria code.

to.etc.domui/src/main/java/to/etc/domui/component/meta/init/MetaInitializer.java
millevlada: actually, seems that this change is no longer needed... I'll revert it, since i already added fix on other place closer to exists subquery

fjalvingh (approved):

PR #100, opened 2020-04-24
Added support for -- inline comments

millevlada:


common/to.etc.alg/src/main/java/to/etc/lexer/ReaderTokenizerBase.java
```diff
@@ -246,4 +265,10 @@ else if(!Character.isWhitespace((char) c))
 	public void setScanUndottedNumbers(boolean scanUndottedNumbers) {
 		m_scanUndottedNumbers = scanUndottedNumbers;
 	}
+
+	public int laNextNonWs() throws IOException {
```
millevlada, line 269: not sure if this might be useful to keep, or I should remove it?

fjalvingh (approved):

PR #101, opened 2020-05-18
Add test ids to Tabs.

stefankaticec:


common/to.etc.alg/src/main/java/to/etc/util/StringTool.java
```diff
@@ -2936,6 +2936,30 @@ public static String strCapitalized(@NonNull String name) {
 		return Character.toUpperCase(c) + name.substring(1).toLowerCase();
 	}
 
+	@NonNull
+	public static String strCapitalizedIntact(@NonNull String name) {
```
fjalvingh, line 2946: Needs Javadoc.

common/to.etc.alg/src/main/java/to/etc/util/StringTool.java
```diff
@@ -2936,6 +2936,30 @@ public static String strCapitalized(@NonNull String name) {
...
+		if(name.length() == 0)
+			return name;
+		char c = name.charAt(0);
+		return Character.toUpperCase(c) + name.substring(1);
+	}
+
+	@NonNull
+	public static String strDecapitlized(@NonNull String name) {
```
fjalvingh: Spelling error in the name.

common/to.etc.alg/src/main/java/to/etc/util/StringTool.java
```diff
@@ -2936,6 +2936,30 @@ public static String strCapitalized(@NonNull String name) {
...
+		if(name.length() == 0)
+			return name;
+		char c = name.charAt(0);
+		return Character.toLowerCase(c) + name.substring(1).toLowerCase();
+	}
+
+	@NonNull
+	public static String strDecapitlizedIntact(@NonNull String name) {
```
fjalvingh: Ditto (decapitalized).

integrations/to.etc.domui.selenium/src/main/java/nl/skarp/DefinitionMgmtListPagePO.java
```diff
@@ -0,0 +1,82 @@
+package nl.skarp;
```
fjalvingh: Really does not belong here...

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/base/AbstractTextPO.java
```diff
@@ -0,0 +1,29 @@
+package to.etc.domui.webdriver.core.base;
```
fjalvingh: base as a class is very... uninformative... These are the component proxies for testing, so testproxies?

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/base/AcePO.java
```diff
@@ -0,0 +1,45 @@
...
+
+	private String getActualId() {
+		var editor = wd().findElement(m_testId);
+		if(editor == null) {
+			throw new IllegalStateException("Can't find Ace Editor with testid: "+ m_testId);
+		}
+		return editor.getAttribute("id");
+	}
+
+	protected JavascriptExecutor jsExecutor() {
```
fjalvingh: Why protected? Should be in basePO?

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/base/BasePO.java
```diff
@@ -0,0 +1,21 @@
+package to.etc.domui.webdriver.core.base;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import to.etc.domui.webdriver.core.WebDriverConnector;
+
+@NonNullByDefault
+public abstract class BasePO {
+	private final WebDriverConnector m_wd;
+
+	public BasePO(WebDriverConnector wd) {
```
fjalvingh: Should accept a testId in the constructor?

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/base/BasePO.java
```diff
@@ -0,0 +1,21 @@
...
+		m_wd = wd;
+	}
+
+	protected WebDriverConnector wd() {
+		return m_wd;
+	}
+
+	protected static String createTestId(String testId) {
```
fjalvingh: Bad method name; this creates the testid selector, not the test id itself.

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/base/BasePagePO.java
```diff
@@ -0,0 +1,64 @@
...
+		return body.getAttribute("pagename").equalsIgnoreCase(getClazz().getName());
+	}
+
+	public Map<String, String> getPageParameters() throws Exception{
+		wd().driver().getCurrentUrl();
+		var list = URLEncodedUtils.parse(new URI(wd().driver().getCurrentUrl()), StandardCharsets.UTF_8);
+		return list.stream().collect(toMap(NameValuePair::getName, NameValuePair::getValue));
+	}
+
+	public void waitUntilPageContains(String part) {
```
fjalvingh: Bad name: it waits for the page URL, not anything in the page. These are also dangerous because you pass in a string; if page classes are refactored, what will happen with these?
As the goal is to test DomUI pages with this, why not pass in Class<? extends UrlPage>?

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/base/BasePagePO.java
```diff
@@ -0,0 +1,64 @@
...
+		var wait = new WebDriverWait(wd().driver(), wd().getWaitInterval());
+		wait.until(ExpectedConditions.urlContains(part));
+	}
+
+	public void waitUntilPageDoesNotContain(String part) {
```
fjalvingh: Ditto.

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/base/BasePagePO.java
```diff
@@ -0,0 +1,64 @@
...
+		var wait = new WebDriverWait(wd().driver(), wd().getWaitInterval());
+		wait.until(new UrlDoesNotContain(part));
+	}
+
+	public String getCurrentUrl() {
+		return wd().driver().getCurrentUrl();
+	}
+
+	protected Class<T> getClazz() {
```
fjalvingh: Method name does not mean anything.

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/base/ButtonPO.java
```diff
@@ -0,0 +1,32 @@
...
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.openqa.selenium.By;
+import to.etc.domui.webdriver.core.WebDriverConnector;
+
+@NonNullByDefault
+public class ButtonPO extends BasePO {
+
+	private final By m_locator;
+
+	public ButtonPO(WebDriverConnector wd, String id) {
```
fjalvingh: Should be testId. You need to have a common contract for these classes.

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/base/BasePO.java
```diff
@@ -0,0 +1,21 @@
+package to.etc.domui.webdriver.core.base;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import to.etc.domui.webdriver.core.WebDriverConnector;
+
+@NonNullByDefault
+public abstract class BasePO {
```
fjalvingh: Test proxies for IControl<T> components should have a common interface, with common methods that are possible on them. like isDisabled() and perhaps even get/setValue.

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/base/DataTablePO.java
```diff
@@ -0,0 +1,91 @@
...
+		private WebDriverConnector m_wd;
+
+		public ColumnPO(WebElement x, WebDriverConnector wd) {
+			m_webElement = x;
+			m_wd = wd;
+		}
+		public void click() {
+			m_webElement.click();
+		}
+		public void check() {
```
fjalvingh: Does not belong here and seems wrong, as this should be a "td", not a checkbox.

fjalvingh: I am actually not fond of this DataTable interface, because for me it's too low level. We should not do too much here until we have a better model in mind. My biggest gripe is with the index based access to columns and rows. That is error prone, I'd rather try something else.

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/base/DataTablePO.java
```diff
@@ -0,0 +1,91 @@
...
+		public ColumnPO getColumn(int column) {
+			var columns =  getColumns();
+			return columns.get(column);
+		}
+
+		public List<ColumnPO> getColumns() {
+			return m_webElement.findElements(
+				By.cssSelector("td")
+			).stream().map(x->new ColumnPO(x,m_wd)).collect(Collectors.toUnmodifiableList());
+		}
```
fjalvingh: This PO actually needs a click() method.

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/base/Either.java
```diff
@@ -0,0 +1,70 @@
+package to.etc.domui.webdriver.core.base;
+
+import org.eclipse.jdt.annotation.Nullable;
+import to.etc.function.ConsumerEx;
+
+public class Either<RIGHT, WRONG> {
```
fjalvingh: Hate this. The name is definitely wrong; either just means one or the other, and implies nothing about right and wrong.
Looking at its usage I much, much prefer just erroring out in normal code.

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/base/ITAbstractUITest.java
```diff
@@ -0,0 +1,9 @@
+package to.etc.domui.webdriver.core.base;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import to.etc.domui.webdriver.core.AbstractWebDriverTest;
+
+@NonNullByDefault
+public class ITAbstractUITest extends AbstractWebDriverTest {
```
fjalvingh: Never create a base class if you have nothing to put in there. They tend to remain forever, and it never saves any reasonable time- it is quick enough to add them later.

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/base/InputPO.java
```diff
@@ -0,0 +1,12 @@
+package to.etc.domui.webdriver.core.base;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.openqa.selenium.By;
+import to.etc.domui.webdriver.core.WebDriverConnector;
+
+@NonNullByDefault
+public class InputPO extends AbstractTextPO {
```
fjalvingh: I noticed we have this one, but not a Text2PO / TextPO... Those should be there (especially for Text2) because their behavior is more complex.

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/base/LookupPO.java
```diff
@@ -0,0 +1,53 @@
+package to.etc.domui.webdriver.core.base;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.openqa.selenium.By;
+import org.openqa.selenium.support.ui.ExpectedConditions;
+import org.openqa.selenium.support.ui.WebDriverWait;
+import to.etc.domui.webdriver.core.WebDriverConnector;
+
+@NonNullByDefault
+public class LookupPO extends BasePO {
```
fjalvingh: Should be LookupInputPO?

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/base/LookupPO.java
```diff
@@ -0,0 +1,53 @@
...
+
+	private final String m_id;
+
+	public LookupPO(WebDriverConnector connector, String id) {
+		super(connector);
+		m_id = id;
+	}
+
+	public void setValue(String val, String colName) {
```
fjalvingh: As discussed, we need to change these. We can do that later, but it needs to be done right after to prevent fragile tests.

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/base/LookupPO.java
```diff
@@ -0,0 +1,53 @@
...
+		}
+		wd().cmd().type(val).on(inputLocator);
+		wd().cmd().click().on(By.cssSelector(".ui-lui2-dlg *[testid='searchButton']"));
+		wd().cmd().click().on(By.cssSelector(".ui-lui2-dlg table td:first-child"));
+		wait.until(ExpectedConditions.invisibilityOfElementLocated(By.cssSelector(".ui-lui2-dlg")));
+	}
+
+
+	public String getValue() {
+		return wd().getElement(By.cssSelector("[testid='" + m_id + "'] .ui-lui-vcell")).getText();
```
fjalvingh: Should actually return something like the selected record's id, I think.

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/base/LookupPO.java
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	public boolean isDisabled() {
```
fjalvingh: That is nice, we need this for the other components too ;)

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/base/LookupPO.java
```diff
@@ -0,0 +1,53 @@
...
+		var locator = wd().byId(m_id + "-lookup");
+		try {
+			var disabled = wd().getAttribute(locator, "disabled");
+			return "true".equalsIgnoreCase(disabled);
+		} catch(IllegalStateException e) {
+			if(e.getMessage().contains("No value for expected attribute disabled at locator")) {
```
fjalvingh: Fragile. We should add "findAttribute()" to WebDriverConnector and let it handle the case where the attribute does not exist, and let it return null in that case. That leaves this problem at one location instead of everywhere.

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/base/MsgBoxPO.java
```diff
@@ -0,0 +1,28 @@
...
+	public String getText() {
+		var el = wd().findElement(createLocator());
+		if(el == null) {
+			throw new IllegalStateException("Can't find MessageBox with id" + m_id);
+		}
+		return el.getText();
+	}
+
+	private By createLocator() {
+		return By.cssSelector("*[testid='" + m_id + "'] .ui-mbx-mc");
```
fjalvingh: Should use that BasePO.createTestId method.

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/base/MsgBoxPO.java
```diff
@@ -0,0 +1,28 @@
...
+	}
+}
```
fjalvingh: We miss a method to actually click the thing, or to click it away.

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/base/RadioButtonPO.java
```diff
@@ -0,0 +1,34 @@
...
+@NonNullByDefault
+public class RadioButtonPO extends BasePO {
+	private final By m_locator;
+
+	public RadioButtonPO(WebDriverConnector wd, String id) {
+		super(wd);
+		m_locator = wd.byId(id);
+	}
+
+	public void select(int order) {
```
fjalvingh: All methods that invoke an action are called click() so far, so why not call this click()?

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/base/RadioButtonPO.java
```diff
@@ -0,0 +1,34 @@
...
+
+	private WebElement getRatioButton() {
+		var rb = wd().findElement(m_locator);
+		if(rb == null) {
+			throw new IllegalStateException("Radio button with " + m_locator + " couldn't be found");
+		}
+		return rb;
+	}
+
+	public void selectWithLabel(String label) {
```
fjalvingh: Should also be called click, and can we have one that accepts Enum<T>? Because 99.9% of the time these are enums.

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/base/SelectPO.java
```diff
@@ -0,0 +1,27 @@
+package to.etc.domui.webdriver.core.base;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.openqa.selenium.By;
+import to.etc.domui.webdriver.core.WebDriverConnector;
+
+@NonNullByDefault
+public class SelectPO extends BasePO {
```
fjalvingh: Should not really exist by itself? I.e. we would need ComboLookup2PO and ilk which could use this as a base/part, of course..

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/base/SelectPO.java
```diff
@@ -0,0 +1,27 @@
...
+
+	private final String m_locator;
+
+	public SelectPO(WebDriverConnector connector, String locator) {
+		super(connector);
+		m_locator = locator;
+	}
+
+	public void select(int idx) {
```
fjalvingh: These need some work too.. We should try to add some more typeful ones.

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/base/TabPO.java
```diff
@@ -0,0 +1,88 @@
...
+import org.openqa.selenium.By;
+import to.etc.domui.webdriver.core.WebDriverConnector;
+import to.etc.util.Pair;
+
+import java.util.Map;
+
+import static java.util.stream.Collectors.toMap;
+
+@NonNullByDefault
+public class TabPO extends BasePO {
```
fjalvingh: TabPanelPO.

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/base/TabPO.java
```diff
@@ -0,0 +1,88 @@
...
+	 */
+	public TabPO(WebDriverConnector wd, String testId, Map<String,? extends BasePO> tabs) {
+		super(wd);
+		m_id = testId;
+		m_tabs = tabs;
+	}
+
+	/**
+	 * Opens a tab with corresponding testID.
+	 * @param testId of the tab
```
fjalvingh: Do not javadoc obvious parameters and results ;)

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/base/TabPO.java
```diff
@@ -0,0 +1,88 @@
...
+	private String m_id;
+
+	private Map<String, ? extends BasePO> m_tabs;
+
+	/**
+	 * @param wd web driver
+	 * @param testId of the Tab
+	 * @param tabs list of configured tabs.
+	 */
+	public TabPO(WebDriverConnector wd, String testId, Map<String,? extends BasePO> tabs) {
```
fjalvingh: If you want to define the tab panels as PO's themselves it's better to do that with an add() method instead of via the constructor, I think.

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/base/TabPO.java
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	/**
+	 * Gets the tabs that are on the page, by provided testId
+	 * @returns a map, where key is the testId of the tab and value is the label.
+	 */
+	public Map<String, String> getTabs() {
+		var elements = wd().findElements(By.cssSelector(createTestId(m_id) + " .ui-tab-hdr li"));
+		return elements.stream().collect(toMap(x->x.getAttribute("testId"), x->x.getText()));
+	}
```
fjalvingh: This model for the TabPanel is very confusing. It has a map of tabs, but also methods that use the actual remote state as the definition for the map of tabs. When to use what is totally unclear...

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/pogeneration/JavaPageObjectPrinter.kt
```diff
@@ -0,0 +1,157 @@
...
+			.replace("(", "")
+			.replace(")", "")
+			.replace("*", "")
+			.replace(";", "")
+			.replace("/", "")
+			.replace("\\", "");
+	}
+
+	private fun appendClassDeclaration() {
+		sb.append("@NonNullByDefault() \n")
```
fjalvingh: Space before \n.

integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/core/pogeneration/JavaPageObjectPrinter.kt
```diff
@@ -0,0 +1,157 @@
...
+
+		return sb.toString()
+	}
+
+	private fun actualClassName(): String {
+		return "${m_actual.name}PO"
+	}
+
+	private fun appendConstructor() {
+		sb.append("\tpublic ${actualClassName()}(WebDriverConnector wd) { \n \tsuper(wd); \n\t }")
```
fjalvingh: There are spaces before the \n and the \t. This means any reformat of this code will bring changes.

to.etc.domui/src/main/java/to/etc/domui/dom/html/Page.java
```diff
@@ -584,6 +584,10 @@ public String	allocateTestID(@NonNull String initial) {
 		return initial + "_" +v;
 	}
 
+	public void dealocateTestId(@NonNull String testId) {
```
fjalvingh: deallocate.. We need to discuss the need of this.

fjalvingh (changes requested):
That was a biggy.... Let's discuss on how to get this in, in parts.

integrations/to.etc.domui.selenium/src/main/java/nl/skarp/DefinitionMgmtListPagePO.java
stefankaticec: I don't see this file? o.O

PR #110, opened 2020-06-26
Added support for reusable bigdecimal number converters

millevlada:


to.etc.domui/src/main/java/to/etc/domui/converter/NumberConverterFactory.java
```diff
@@ -48,4 +57,23 @@ public int accept(Class< ? > clz, PropertyMetaModel< ? > pmm) {
...
 	}
+
+	/**
+	 * Returns reusable {@link BigDecimalNumberConverter} instance based on specified numeric presentation and scale.
+	 * @return
+	 */
+	static public IConverter<BigDecimal> createBigDecimalNumberConverters(NumericPresentation np, int scale) {
+		Pair<NumericPresentation, Integer> key = new Pair<>(np, Integer.valueOf(scale));
+		NumberConverter<BigDecimal> bdConverter = BIG_DECIMAL_CONVERTERS.get(key);
+		if (null == bdConverter) {
```
fjalvingh: You cannot do this; this is a common pattern that fails. The best way to handle this is to use a ConcurrentHashMap, and to use computeIfAbsent - that will atomically do the right thing.

to.etc.domui/src/main/java/to/etc/domui/converter/NumberConverterFactory.java
```diff
@@ -24,10 +24,19 @@
...
 import to.etc.domui.util.DomUtil;
+import to.etc.util.Pair;
+
+import java.math.BigDecimal;
+import java.util.HashMap;
+import java.util.Map;
 
 public class NumberConverterFactory implements IConverterFactory {
+
+	static Map<Pair<NumericPresentation, Integer>, NumberConverter<BigDecimal>> BIG_DECIMAL_CONVERTERS = new HashMap<>();
```
fjalvingh: private

fjalvingh (changes requested):
MT issue.

fjalvingh (approved):

PR #128, opened 2020-07-31
fa-save from Font-Awesome, file upload clear click handler

millevlada:
Added fa-save from Font-Awesome 4.7.0
Added optional onClearClicked handler to FileUpload2
Fixed few typos in api docs


to.etc.domui/src/main/java/to/etc/domui/component/upload/FileUpload2.java
```diff
@@ -407,5 +426,15 @@ public boolean handleUploadRequest(@NonNull RequestContextImpl param, @NonNull C
 	@Override public void setHint(String hintText) {
 		setTitle(hintText);
 	}
+
+	@Nullable
+	public IClicked<FileUpload2> getOnClearClicked() {
```
fjalvingh, line 431: Why can we not just have an onValueChanged handler? That should see when the value is cleared?

fjalvingh:
Merged into skarp-prd.

PR #152, opened 2020-10-20
Added util to create enum combo with specified converter for ComboFix…

millevlada:
…ed2.

Fixed TabPanel for isStretch use.
Added grid display type member.


integrations/fontawesome4/src/main/java/to/etc/domui/fontawesome/FaIconGenerator.java
```diff
@@ -113,7 +113,7 @@ static String alterName(String faname) {
 		"fa-quote-right", "fa-random", "fa-ravelry", "fa-rebel", "fa-recycle", "fa-reddit", "fa-reddit-alien",
 		"fa-reddit-square", "fa-refresh", "fa-registered", "fa-renren", "fa-repeat", "fa-reply", "fa-reply-all",
 		"fa-retweet", "fa-road", "fa-rocket", "fa-rss", "fa-rss-square", "fa-rub", "fa-safari", "fa-scissors",
-		"fa-scribd", "fa-search", "fa-search-minus", "fa-search-plus", "fa-sellsy", "fa-server", "fa-share",
+		"fa-scribd", "fa-search", "fa-search-minus", "fa-search-plus", "fa-sellsy", "fa-send", "fa-send-o", "fa-server", "fa-share",
```
fjalvingh, line 116: It's really odd that all those things are missing. Perhaps we should try to re-generate these files properly..

fjalvingh (approved):

PR #153, opened 2020-10-21
Fix timer

stefankaticec:


common/to.etc.alg/src/main/java/to/etc/util/TimerUtil.java
```diff
@@ -23,7 +23,7 @@ static private final ScheduledExecutorService getScheduler() {
 			m_scheduler = scheduler = Executors.newScheduledThreadPool(2, new ThreadFactory() {
 				@Override
 				public Thread newThread(Runnable r) {
-					Thread t = new Thread();
+					Thread t = new Thread(r);
```
fjalvingh, line 26: That was terrible 8-/ I will clean the flogging device and apply it sternly this afternoon 8-/

fjalvingh (approved):

PR #167, opened 2020-11-12
Added few more effects animations.

millevlada:


to.etc.domui/src/main/java/to/etc/domui/dom/Animations.java
```diff
@@ -95,4 +95,33 @@ static public void shake(NodeBase node) {
...
+	static public void bounce(NodeBase node) {
+		node.appendStatement()
+			.select(node)
+			.append(".effect('bounce')")
+			.next();
+	}
+
+	/**
+	 * Does pulsate effect.
+	 * @param node
```
fjalvingh, line 107: Please do not add param comments if you leave them empty (I'm doing a lot of work to remove those).

fjalvingh (approved):

PR #178, opened 2020-12-09
Add year end.

stefankaticec:


common/to.etc.alg/src/main/java/to/etc/util/DateUtil.java
```diff
@@ -406,6 +406,18 @@ static public Date yearStartDate(int year) {
 		return new Date(cal.getTimeInMillis());
 	}
 
+	static public Date yearEndDate(int year) {
```
fjalvingh, line 409: This is not a reasonable thing to have. The last date of the year should not have a timestamp, as that timestamp makes it strange: there IS no last time as it depends on the resolution of the time you are working with. Here you set it to 23:59:59.999, but if you use java.sql.Date the resolution can be much higher. To capture all data of a previous year you would just use 0:00:00.000 of the next year and demand everything before it.

What is reasonable is to have the last date (no time) of the year as xxxx-12-31..

fjalvingh (changes requested):

fjalvingh (approved):

PR #194, opened 2021-03-28
Fix control creator enum and bool

gerardv:
I kept running into this at compile time. Don't understand why this happens on my machine and not on yours, or you'd have it fixed already. That, or there's something I'm not doing right, off course. :)

### DIFF
--- a/to.etc.domui/src/main/java/to/etc/domui/component2/controlfactory/ControlCreatorEnumAndBool.java
+++ b/to.etc.domui/src/main/java/to/etc/domui/component2/controlfactory/ControlCreatorEnumAndBool.java
@@ -55,7 +55,7 @@ public class ControlCreatorEnumAndBool implements IControlCreator {
 	@Override
 	public <T, C extends IControl<T>> C createControl(@NonNull PropertyMetaModel<T> pmm, @Nullable Class<C> controlClass) {
 		Object[] domainValues = pmm.getDomainValues();
-		if(domainValues == null ||  domainValues.length > 5 || controlClass == ComboFixed2.class) {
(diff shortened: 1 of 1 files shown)


to.etc.domui/src/main/java/to/etc/domui/component2/controlfactory/ControlCreatorEnumAndBool.java
```diff
@@ -55,7 +55,7 @@
 	@Override
 	public <T, C extends IControl<T>> C createControl(@NonNull PropertyMetaModel<T> pmm, @Nullable Class<C> controlClass) {
 		Object[] domainValues = pmm.getDomainValues();
-		if(domainValues == null ||  domainValues.length > 5 || controlClass == ComboFixed2.class) {
+		if(controlClass != null && (domainValues == null ||  domainValues.length > 5 || controlClass.equals(ComboFixed2.class))) {
```
fjalvingh, line 58: Can you explain why you wanted this? Because the original code seems to be correct. You are using equals() to compare a class. Equals is a "compare by value" thing. I'm using == which compares the class references (i.e. not the class's content but the pointer to the class. If both sides are the same class their pointer is, for classes and some other Java constructs, also the same).

gerardv: Yeah, you're right. I bent my head over it and this was the only thing that a) I was able to come up with while actually trying to accomplish something completely different and b) managed to get it to compile... and that's where the actual answer was: the compiler. Somehow it switched back to javac. I set it to Eclipse again and it compiled without errors now. For the record: the error I got was:
`java: incomparable types: java.lang.Class<C> and java.lang.Class<to.etc.domui.component2.combo.ComboFixed2>`

PR #204, opened 2021-04-08
add an INFO type MsgBox with an IAnswer parameter.

gerardv:
I need a callback after displaying a TYPE.info msgbox to the user. This allows for a forceRebuild() of the calling page without losing the msgbox altogether, for example.

### DIFF
--- a/to.etc.domui/src/main/java/to/etc/domui/component/misc/MsgBox.java
+++ b/to.etc.domui/src/main/java/to/etc/domui/component/misc/MsgBox.java
@@ -287,6 +287,16 @@ public class MsgBox extends Window {
 		message(dad, Type.INFO, string);
 	}
 
+	/**
+	 * Display some information; Call the onAnswer handler when 'Ok' is clicked.
+	 * @param dad
+	 * @param string
+	 * @param onAnswer
+	 */
+	public static void info(NodeBase dad, String string, final IAnswer onAnswer) {
+		message(dad, MsgBox.Type.INFO, string, onAnswer);
+	}
+
 	public static void warning(NodeBase dad, NodeContainer string) {
 		message(dad, Type.WARNING, string);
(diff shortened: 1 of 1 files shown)


to.etc.domui/src/main/java/to/etc/domui/component/misc/MsgBox.java
```diff
@@ -287,6 +287,16 @@ public static void info(NodeBase dad, NodeContainer string) {
 		message(dad, Type.INFO, string);
 	}
 
+	/**
```
fjalvingh, line 290: Adding Javadoc is really good, but do not add those empty @param tags. We only add those (and the @result and @throws too) if they really need explanation. You can disable the automatic generation of those things in IntelliJ.

to.etc.domui/src/main/java/to/etc/domui/component/misc/MsgBox.java
```diff
@@ -287,6 +287,16 @@ public static void info(NodeBase dad, NodeContainer string) {
...
+	 * Display some information; Call the onAnswer handler when 'Ok' is clicked.
+	 * @param dad
+	 * @param string
+	 * @param onAnswer
+	 */
+	public static void info(NodeBase dad, String string, final IAnswer onAnswer) {
```
fjalvingh, line 296: We no longer use final in those parameters. It makes the already very verbose Java code even more verbose, and does not really add value. I personally have no problem with people changing parameter values anyway.

fjalvingh: Btw, perhaps consider using MsgBox2?
```
MsgBox2.on(this).info().onanswer(....).
```

fjalvingh (approved):

PR #210, opened 2021-04-13
Fixed margin for first button in dialog

millevlada:


to.etc.domui/src/main/java/to/etc/domui/util/asyncdialog/AsyncDialog.java
```diff
@@ -44,24 +62,33 @@ private AsyncDialog() {
 						}
 					}
 				} else {
+					if(null == errorException && cancelled) {
```
fjalvingh: That is wrong. If you cancel an operation but it finishes before it can be cancelled then it is finished, not cancelled.
A cancelled operation MUST by itself throw the cancellation exception if it cancels.

fjalvingh: And those ifs are really messy, but I think they will mostly disappear again.

to.etc.domui/src/main/java/to/etc/domui/util/Msgs.java
```diff
@@ -151,6 +151,9 @@
 	/** Value out of range */
 	public static String V_OUT_OF_RANGE = "v.out.of.range";
 
+	/** Invalid not regex */
+	public static String V_INVALID_REGEX = "v.invalid.regex";
```
fjalvingh: Why would that error be here? It is nowhere used within DomUI.
It should also be an enum constant if we do need it.

to.etc.domui/src/main/resources/resources/themes/scss/winter/_buttonBar.scss
```diff
@@ -11,6 +11,9 @@
 .ui-bb-middle .ui-button {
 	margin: 5px 10px 5px 0;
 }
+.ui-bb-middle .ui-button:first-of-type {
+	margin-left: 10px;
```
fjalvingh: What exactly is solved by this? Can you show a screenshot? The button bar is used really a lot so changes are a bit dangerous.

to.etc.domui/src/main/resources/resources/themes/scss/winter/_floatingWindow.scss
```diff
@@ -72,7 +72,7 @@
 }
 
 .ui-flw-btn-close {
-	padding: 0px;
+	padding: 0px 0px 0px 4px;
```
fjalvingh, line 79: Same for here.

to.etc.domui/src/main/resources/to/etc/domui/util/messages.properties
```diff
@@ -98,6 +98,7 @@ v.bad.duration=Ongeldige duur (voorbeeld: 12d 8h)
 v.bad.percentage=Ongeldig percentage ({0})
 v.invalid.operator=Ongeldige zoek-operatie ({0}) in zoekveld
 v.invalid.not.unique=Waarde ({0}) al in gebruik
+v.invalid.regex=Falsches Regex-Muster: {0}, waarde ({1})
```
fjalvingh: If we need it, should be: Ongeldige regular expression: {0}, waarde {1}

fjalvingh (changes requested):

to.etc.domui/src/main/java/to/etc/domui/util/asyncdialog/AsyncDialog.java
millevlada: actually issue that I wanted to solve was that if I cancel, and it was not even started, I have no distinction from case when it successfully completed. I think if it completes  before it handles the cancel, it actually completes, and that cancelled here is false.

to.etc.domui/src/main/java/to/etc/domui/util/Msgs.java
millevlada: well, why not? I mean, isn't it convenient to have it here? There are also other validation messages that we have, that seem unused in domui as well? I see that V_INVALID_DOUBLE and V_OUT_OF_RANGE are also unused?

to.etc.domui/src/main/resources/resources/themes/scss/winter/_buttonBar.scss
millevlada: well, if we have more width in buttons then in content then first button gets sticked to left border of dialog, I had that reported as issue actually.

to.etc.domui/src/main/resources/resources/themes/scss/winter/_floatingWindow.scss
millevlada: looks ugly ;) Title gets sticked to close button otherwise, when title has more width than content.

to.etc.domui/src/main/resources/to/etc/domui/util/messages.properties
millevlada: Ok, I saw that translation already in use in code in some sql tool... Ah I see, it was _de_ . Ok, I'll fix that to your proposed text.

to.etc.domui/src/main/java/to/etc/domui/util/Msgs.java
fjalvingh: Because others did something wrong I'm now allowed too is never going to fly well..

It is one thing to have really common messages here. But this one is really not common...

You can leave it, but do change it to an enum field (I need to do that with all others too, and it is a shitload of work).

millevlada: Well, in my eyes it was not wrong at all. It is not that I did it because others did it before -> I did it because I think it is useful to have it as one of possible built in validation messages. But ok... So, you want me to spend a day in moving all that to an enum? It has 721 usages, everywhere... And it would also break the backward compatibility of domui used elsewhere. Do we really want that now?

to.etc.domui/src/main/resources/resources/themes/scss/winter/_buttonBar.scss
millevlada: re-fixed to be used only within floating div style

to.etc.domui/src/main/resources/to/etc/domui/util/messages.properties
millevlada: fixed translation

to.etc.domui/src/main/java/to/etc/domui/util/Msgs.java
millevlada: ah, ok, you meant to add that new entry as enum member... Wasn't aware that we have that partially done already here... Ofc, fixed.

to.etc.domui/src/main/java/to/etc/domui/util/asyncdialog/AsyncDialog.java
millevlada: Agreed to temporarily leave this change out. We need different solution for this.

fjalvingh (approved):

PR #215, opened 2021-05-05
Added change of order of handling the access checks in domui page

millevlada:


millevlada:
decided to do it different

PR #217, opened 2021-05-06
Access check injection

millevlada:


millevlada:
closed, we choose different solution, with typed exception

PR #220, opened 2021-05-17
Security utils more

millevlada:


common/to.etc.alg/src/main/java/to/etc/util/SecurityUtils.java
```diff
@@ -52,7 +52,7 @@
 
 	static private final SecureRandom RANDOM = new SecureRandom();
 
-	public static final String ENCRYPTION_CIPHER = "Blowfish";
+	private enum Algorythm { Blowfish, AES }
```
fjalvingh: Algorithm

common/to.etc.alg/pom.xml
```diff
@@ -51,7 +51,10 @@
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
         </dependency>
-
+        <dependency>
```
fjalvingh: We do not need this.

common/to.etc.alg/src/main/java/to/etc/util/SecurityUtils.java
```diff
@@ -283,29 +283,49 @@ static public boolean checkPassword(@NonNull String encodedPassword, @NonNull St
...
+	}
+
+	static public String decryptStringAes(String input, String password) throws Exception {
+		byte[] bytes = StringTool.decodeBase64(input);
+		return decryptStringFromBytes(bytes, password, Algorythm.AES);
+	}
+
+	static public String encryptStringAesHex(String input, String password) throws Exception {
+		byte[] bytes = encryptStringToBytes(input, password, Algorythm.AES);
+		return Hex.encodeHexString(bytes);
```
fjalvingh: StringTool.toHex()

common/to.etc.alg/src/main/java/to/etc/util/SecurityUtils.java
```diff
@@ -283,29 +283,49 @@ static public boolean checkPassword(@NonNull String encodedPassword, @NonNull St
...
+	}
+
+	static public String decryptStringAesHex(String input, String password) throws Exception {
+		byte[] bytes = Hex.decodeHex(input.toCharArray());
```
fjalvingh: StringTool.fromHex

common/to.etc.alg/src/main/java/to/etc/util/SecurityUtils.java
```diff
@@ -283,29 +283,49 @@ static public boolean checkPassword(@NonNull String encodedPassword, @NonNull St
...
 		return StringTool.encodeBase64ToString(bytes);
 	}
 
 	static public String decryptString(String input, String password) throws Exception {
 		byte[] bytes = StringTool.decodeBase64(input);
-		return decryptStringFromBytes(bytes, password);
+		return decryptStringFromBytes(bytes, password, Algorythm.Blowfish);
+	}
+
+	static public String encryptStringAes(String input, String password) throws Exception {
```
fjalvingh: Why not just have base64 and hex variants and pass in the algorithm? We now have to add a whole set of methods for every algorithm added...

fjalvingh (changes requested):

common/to.etc.alg/pom.xml
millevlada: ? It has handy methods that I use to encode to hex and to decode from hex. Why we should not use it?

common/to.etc.alg/src/main/java/to/etc/util/SecurityUtils.java
millevlada: we had encryptString and decryptString in old code base that used Blowfish and base64. So, I had to keep it for backward compatibility. That is why I just added 2 more pairs of 2 lines methods, that I used in my code, with reusing the stuff that was there already. Isn't it better to have dedicated methods then combining the parameters in each call? I mean, can be easier to make the mistake.

common/to.etc.alg/src/main/java/to/etc/util/SecurityUtils.java
millevlada: ok, I see, that is why we do not need that library... Ok, I'll switch to that.

common/to.etc.alg/pom.xml
millevlada: ok, I see we have that StringTool. I just saw that Hex util used in other code, but I agree we do not need to introduce it here. Fixed.

common/to.etc.alg/src/main/java/to/etc/util/SecurityUtils.java
millevlada: fixed

common/to.etc.alg/src/main/java/to/etc/util/SecurityUtils.java
millevlada: ok, I made it as asked.

common/to.etc.alg/src/main/java/to/etc/util/SecurityUtils.java
millevlada: fixed

fjalvingh (approved):

PR #223, opened 2021-05-24
Added support for navigation checks based on modified and unsaved dat…

millevlada:
…a on pages.


to.etc.domui/src/main/java/to/etc/domui/dom/html/IPageWithNavigationCheck.java
```diff
@@ -0,0 +1,19 @@
...
+	 * Returns if page has modifications that should prevent navigation.
+	 * @return
+	 */
+	boolean hasModification();
+
+	/**
+	 * Call to let user handle how to resolve the navigation on modified screen data.
+	 * @return
+	 */
+	void handleNavigationOnModified();
```
fjalvingh: These two things should not be combined. The using of the interface should force some standard mechanism (in both Javascript for back button and DomUI for DomUI level changes). If you want to control what DomUI action to take it needs to be a something registered in DomApplication which can be overridden, and optionally some other method/interface in the page. That way the detecting of the modified (and the signal to use that) and the handling of that occurrence are not coupled.

to.etc.domui/src/main/java/to/etc/domui/dom/html/Page.java
```diff
@@ -1033,6 +1035,14 @@ void registerJavascriptStateChanged(@NonNull NodeBase nodeBase) {
 	 */
 	@Nullable
 	public StringBuilder internalFlushJavascriptStateChanges() throws Exception {
+		if(canLeaveCurrentPage(false)) {
+			if(m_rootContent instanceof IPageWithNavigationCheck) {
+				m_rootContent.appendJavascript("window.onbeforeunload = null;");
+			}
+		}else {
+			//seems to be hardcoded message in browsers, so text that we return here is ignored.
+			m_rootContent.appendJavascript("window.onbeforeunload = function() { return \"You have unsaved changes on page.\"};");
```
fjalvingh: That javascript should be a WebUI function so that we can easily change it. It also seems incomplete: the web states that you need to do:
```
window.addEventListener('beforeunload', (event) => {
            event.preventDefault();
           // Google Chrome requires returnValue to be set.
            event.returnValue = '';
        });
```
Returning a user message is not supported anymore for security.

to.etc.domui/src/main/java/to/etc/domui/dom/html/Page.java
```diff
@@ -1380,4 +1390,31 @@ public void execute(@NonNull IExecute code) {
 				page.addDelayedExecution(code);
 		}
 	}
+
+	/**
+	 * Checks if page can be left.
+	 * @return
+	 */
+	public boolean canLeaveCurrentPage(boolean handleNavigationWhenModified) {
```
fjalvingh: This should not be called by any user code. Name it internalCanLeaveCurrentPage().

to.etc.domui/src/main/java/to/etc/domui/dom/html/Page.java
```diff
@@ -1380,4 +1390,31 @@ public void execute(@NonNull IExecute code) {
...
+			IPageWithNavigationCheck pageWithNavigationCheck = (IPageWithNavigationCheck) m_rootContent;
+			boolean hasModification = pageWithNavigationCheck.hasModification();
+			if(hasModification && handleNavigationWhenModified) {
+				pageWithNavigationCheck.handleNavigationOnModified();
+			}
+			return !hasModification;
+		}else {
+			return true;
+		}
+	}
```
fjalvingh, line 1405: Ok, I do not understand how this works.

I do understand that the handler code needs to call that "setForceLeavePage" method to actually effect a leave. But how does it know where to leave to? I.e. where is the original "target" of the UIGoto action kept?

to.etc.domui/src/main/java/to/etc/domui/dom/html/Page.java
```diff
@@ -1380,4 +1390,31 @@ public void execute(@NonNull IExecute code) {
...
+
+	/**
+	 * Forces page leave flag, it would ignore page leave check on url page level.
+	 */
+	public void setForceLeavePage() {
```
fjalvingh: This is an obscure way to handle this: no one would be able to really find this as part of that interface.

What is a better (and self explanatory) way of doing this is by having the handler of the "the page is modified" action receive a parameter that contains the method to be called. That might also be the place where to store the missing UIGoto information..

to.etc.domui/src/main/java/to/etc/domui/dom/html/Page.java
millevlada: Look at the skarp code pair, since that is the consumer part of this code, I think that all would be clear then.

to.etc.domui/src/main/java/to/etc/domui/dom/html/IPageWithNavigationCheck.java
millevlada: as agreed, splitted into 2 interfaces and also default implementation in dom application.

to.etc.domui/src/main/java/to/etc/domui/dom/html/Page.java
millevlada: as agreed, made it to use WebUI functions.

to.etc.domui/src/main/java/to/etc/domui/dom/html/Page.java
millevlada: named it internal

to.etc.domui/src/main/java/to/etc/domui/dom/html/Page.java
millevlada: changed it, now we have UIGotoContext and real postponed navigation to target of original postponed navigation

to.etc.domui/src/main/java/to/etc/domui/dom/html/Page.java
millevlada: removed it, not needed now

to.etc.domui/src/main/java/to/etc/domui/dom/html/IPageWithDomuiNavigationCheck.java
```diff
@@ -0,0 +1,16 @@
...
+import to.etc.domui.state.UIGotoContext;
+
+/**
+ * Marks the UrlPage for custom handling of leaving the page with unsaved data, caused by domui navigation.
+ */
+public interface IPageWithDomuiNavigationCheck extends IPageWithNavigationCheck {
+
+	/**
+	 * Implements page specific code how user can handle navigation on modified screen data.
+	 * @return
```
fjalvingh, line 13: Try not to add unused javadoc things. You can switch off generation of those in IntelliJ.

to.etc.domui/src/main/java/to/etc/domui/dom/html/Page.java
```diff
@@ -875,6 +876,14 @@ public void appendJS(@NonNull final CharSequence sq) {
 
 	@Nullable
 	public StringBuilder internalFlushAppendJS() {
+		if(internalCanLeaveCurrentPageByBrowser()) {
```
fjalvingh, line 879: These checks need to be the other way around. There is no need to do anything if that interface is not there.

to.etc.domui/src/main/java/to/etc/domui/dom/html/IPageWithDomuiNavigationCheck.java
```diff
@@ -0,0 +1,16 @@
+package to.etc.domui.dom.html;
+
+import org.eclipse.jdt.annotation.NonNull;
+import to.etc.domui.state.UIGotoContext;
+
+/**
+ * Marks the UrlPage for custom handling of leaving the page with unsaved data, caused by domui navigation.
+ */
+public interface IPageWithDomuiNavigationCheck extends IPageWithNavigationCheck {
```
fjalvingh, line 9: I propose to rename that to IPageWithNavigationHandler, because it better specifies what it does. I can do that next merge, no need to do it here.

fjalvingh (approved):

PR #258, opened 2021-07-09
Fixing reasonable default for decimal input controls in form builders.

millevlada:


fjalvingh:
Was merged in some other fix.

PR #277, opened 2021-08-31
Added simple interface for objects that can store and load its state.

millevlada:


to.etc.domui/src/main/java/to/etc/domui/state/IWithStoredState.java
```diff
@@ -0,0 +1,10 @@
+package to.etc.domui.state;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+
+@NonNullByDefault
+public interface IWithStoredState {
```
fjalvingh: We do not use Object as parameter type anymore. Use <T> on the interface and let T replace Object.

fjalvingh (changes requested):

fjalvingh (approved):

PR #279, opened 2021-09-08
Added diagnostic util as handy tool to log performance of certain loo…

millevlada:
…ped code blocks in cumulative way.

### DIFF
--- a/common/to.etc.alg/src/main/java/to/etc/util/DiagnosticUtil.java
+++ b/common/to.etc.alg/src/main/java/to/etc/util/DiagnosticUtil.java
@@ -1,0 +1,144 @@
+package to.etc.util;
+
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Handy util to do profiling on duration spent in certain code blocks.
+ * Call logInit to initialize in memory logging that is stored in thread locale.
+ * Call logStart with given string key-name on start of code block to measure. Call logStop with given same key-name at end of code block, to stop measuring.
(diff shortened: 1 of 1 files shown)


common/to.etc.alg/src/main/java/to/etc/util/DiagnosticUtil.java
```diff
@@ -0,0 +1,144 @@
...
+ * test iteration duration_#1 started at 13:29:26.759=11
+ * test iteration duration_#2 started at 13:29:26.774=15
+ * test iteration duration_#3 started at 13:29:26.790=15
+ *
+ * To get the Map with durations report, call report method at the end, with specified time unit for reporting. That also unregister logging from thread locale
+ * automatically.
+ *
+ * If initLog is not called, called log methods are ignored, and report returns null result.
+ */
+public class DiagnosticUtil {
```
fjalvingh, line 29: Should be final, should have a private constructor.

common/to.etc.alg/src/main/java/to/etc/util/DiagnosticUtil.java
```diff
@@ -0,0 +1,144 @@
...
+
+	private static ThreadLocal<Map<String, Object>> diagnostic = new ThreadLocal<>();
+	private static ThreadLocal<Boolean> diagnosticEnabled = new ThreadLocal<>();
+
+	private static final String SUB_MAP = "_subMap";
+
+	private static final SimpleDateFormat DF = new SimpleDateFormat("HH:mm:ss.SSS");
```
fjalvingh, line 36: These are not reentrant, so you cannot use the instance from multiple threads. And this code WILL be called from multiple threads... You have no other choice than to just allocate one when needed.

fjalvingh (approved):

PR #281, opened 2021-09-09
Added BatchUtil for easier bulk JDBC actions on hibernate objects.

millevlada:
Fixed review remarks.

### DIFF
--- a/common/to.etc.alg/src/main/java/to/etc/util/DiagnosticUtil.java
+++ b/common/to.etc.alg/src/main/java/to/etc/util/DiagnosticUtil.java
@@ -1,5 +1,7 @@
 package to.etc.util;
 
+import org.eclipse.jdt.annotation.NonNull;
+
 import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


common/to.etc.alg/src/main/java/to/etc/util/DiagnosticUtil.java
```diff
@@ -26,14 +28,17 @@
...
+public final class DiagnosticUtil {
 
 	private static ThreadLocal<Map<String, Object>> diagnostic = new ThreadLocal<>();
 	private static ThreadLocal<Boolean> diagnosticEnabled = new ThreadLocal<>();
 
 	private static final String SUB_MAP = "_subMap";
 
-	private static final SimpleDateFormat DF = new SimpleDateFormat("HH:mm:ss.SSS");
+	@NonNull
+	private static final ThreadLocal<SimpleDateFormat> DF = new ThreadLocal<>().withInitial(() -> new SimpleDateFormat("HH:mm:ss.SSS"));
```
fjalvingh, line 39: We do not really do this anymore. Creating a SimpleDateFormat is a pity but as long as you do not do it in a loop for 213132241324 times it is pretty cheap. And in other code we use FastDateFormat.
These ThreadLocals are being complained about when Tomcat toggles a server..

to.etc.domui/src/main/java/to/etc/domui/util/db/BatchUtil.java
```diff
@@ -0,0 +1,76 @@
...
+	private BatchUtil() {}
+
+	/**
+	 * Does bulk delete using JDBC. Use it when performance matters ;)
+	 * @throws SQLException
+	 */
+	public static <K extends Number, T extends IIdentifyable<K>> boolean bulkDelete(@NonNull Connection con, ClassMetaModel cmm, @NonNull List<T> items) throws SQLException {
+		final PropertyMetaModel<?> pkPmm = cmm.getPrimaryKey();
+		if(null == pkPmm) {
+			throw new IllegalArgumentException("Unsupported working with class models that do not have PK columns! For " + cmm.getTableName() + " fund no PK!");
```
fjalvingh, line 29: Text is strange (2 times). Should be "Found no PK for " + tablename), and "Unsupported" should be "xxxxx is not supported" or "xxxxxx is unsupported".

fjalvingh (approved):

PR #290, opened 2021-10-07
Fixed string like search to include new lines into .* regex.

millevlada:
Added util for bulk delete with in/ not in list of IDs.

### DIFF
--- a/to.etc.domui/src/main/java/to/etc/domui/util/db/BatchUtil.java
+++ b/to.etc.domui/src/main/java/to/etc/domui/util/db/BatchUtil.java
@@ -45,6 +45,31 @@ final public class BatchUtil {
 	}
 
 	/**
+	 * Does bulk delete using JDBC.
+	 * In case that except is T, it deletes all other records.
+	 * Otherwise it deletes specified records.
+	 * Use it when performance matters ;)
+	 * @throws SQLException
+	 */
+	public static <K extends Number, T extends IIdentifyable<K>> boolean bulkDelete(@NonNull Connection con, ClassMetaModel cmm, boolean except, @NonNull List<Long> ids) throws SQLException {
+		if(ids.isEmpty() && !except) {
+			return false;
+		}
+		final PropertyMetaModel<?> pkPmm = cmm.getPrimaryKey();
+		if(null == pkPmm) {
+			throw new IllegalArgumentException("Found no PK for " + cmm.getTableName() + " !");
+		}
+		String[] pkCols = pkPmm.getColumnNames();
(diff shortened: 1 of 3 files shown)


to.etc.domui/src/main/java/to/etc/domui/util/db/BatchUtil.java
```diff
@@ -44,6 +44,31 @@ private BatchUtil() {}
...
 	}
 
+	/**
+	 * Does bulk delete using JDBC.
+	 * In case that except is T, it deletes all other records.
+	 * Otherwise it deletes specified records.
+	 * Use it when performance matters ;)
+	 * @throws SQLException
+	 */
+	public static <K extends Number, T extends IIdentifyable<K>> boolean bulkDelete(@NonNull Connection con, ClassMetaModel cmm, boolean except, @NonNull List<Long> ids) throws SQLException {
```
fjalvingh: The generics here do not make sense because they are unused. I think the idea would be to have the ids array be of type K instead of Long. And K should be unbounded (keys are very often strings too).
You can then still assume that tostring of such an ID is suitable to be used with setting inside a ps.setString().

to.etc.domui/src/main/java/to/etc/domui/util/db/BatchUtil.java
```diff
@@ -44,6 +44,31 @@ private BatchUtil() {}
...
+			throw new IllegalArgumentException("Found no PK for " + cmm.getTableName() + " !");
+		}
+		String[] pkCols = pkPmm.getColumnNames();
+		if(pkCols.length != 1) {
+			throw new IllegalArgumentException("Table: " + cmm.getTableName() + " has composite PK: " + pkCols + ". Composite PK is not supported!");
+		}
+		String pkCol = pkCols[0];
+		String idsValues = ids.stream().map(it -> "" + it).collect(Collectors.joining(","));
+		String notPart = except ? "not " : "";
+		return JdbcUtil.executeStatement(con, "delete from " + cmm.getTableName() + " where " + notPart + pkCol + " in (" + idsValues + ")");
```
fjalvingh: This allows SQL injection when we also allow strings. You need to use SQL parameters.

fjalvingh (changes requested):

fjalvingh (approved):

to.etc.domui/src/main/java/to/etc/domui/util/db/BatchUtil.java
millevlada: ok, fixed, now we are not bound to any narrow key types

PR #302, opened 2021-11-30
Date fix

buzakovica:


common/to.etc.alg/src/main/java/to/etc/util/DateUtil.java
```diff
@@ -55,14 +55,16 @@
 	@NonNull
 	private final static ThreadLocal<Calendar> m_testDate = new ThreadLocal<>();
 
+	public static final String DATE_MASK = "yyyyMMdd";
```
fjalvingh, line 58: This does not belong here; it does not mean anything. You use it in your Excel fix where it DOES seem to mean something - namely a format Excel uses as its default format. So it belongs there, not here.

to.etc.domui/src/main/java/to/etc/domui/util/importers/ExcelImportRow.java
```diff
@@ -109,7 +115,12 @@ public String getName() {
 						return String.valueOf(m_cell.getBooleanCellValue());
 
 					case NUMERIC:
-						return m_row.m_rr.convertDouble(m_cell.getNumericCellValue());
+						try {
```
fjalvingh, line 118: ? But this is clearly not right. A number is a number, not a date.

fjalvingh (changes requested):
Seems to be incorrect.

PR #304, opened 2021-12-08
Prd stack trc

buzakovica:


to.etc.domui/src/main/java/to/etc/domui/server/AppFilter.java
```diff
@@ -73,6 +73,9 @@
 
 	static private String m_appContext;
 
+	@Nullable
+	static private Boolean m_isProductionMode;
```
fjalvingh: This should not be a Boolean but a boolean, and it is the wrong way around; it should be isTestMode (as that is what it is called everywhere) and it should default to false.

And boolean field names never have "is" in their name. It would create odd setter names.

to.etc.domui/src/main/java/to/etc/domui/server/AppFilter.java
```diff
@@ -90,6 +93,11 @@ static public synchronized void setIoWrapper(@NonNull IRequestResponseWrapper ww
 		m_ioWrapper = ww;
 	}
 
+	@Nullable
+	public static Boolean isProductionMode() {
```
fjalvingh: Should be isTestMode, and most certainly not be nullable- that does not mean anything.

to.etc.domui/src/main/java/to/etc/domui/server/AppFilter.java
```diff
@@ -189,7 +197,7 @@ public synchronized void init(final FilterConfig config) throws ServletException
 
 		initLogConfig(approot, config.getInitParameter("logpath"));
 
-		if(DeveloperOptions.isDeveloperWorkstation() ) {
+		if(DeveloperOptions.isDeveloperWorkstation()) {
```
fjalvingh: Can you check your IntelliJ settings? There is still a lot of formatting committed here, and that makes it difficult to see the real changes. You should set the save actions for automatic formatting to "off" until I have reformatted all of DomUI.

to.etc.domui/src/main/java/to/etc/domui/server/AppFilter.java
```diff
@@ -220,15 +228,20 @@ public synchronized void init(final FilterConfig config) throws ServletException
...
+			String autoload = domUiReload != null ? domUiReload : m_config.getString("auto-reload");            // Allow override of web.xml values.
 
 			//these patterns will be only watched not really reloaded. It makes sure the reloader kicks in. Found bundles and MetaData will be reloaded only.
 			String autoloadWatchOnly = m_config.getString("auto-reload-watch-only");
 
-			if(DeveloperOptions.isDeveloperWorkstation() && DeveloperOptions.getBool("domui.developer", true) && autoload != null && autoload.trim().length() > 0)
+			if(DeveloperOptions.isDeveloperWorkstation() && DeveloperOptions.getBool("domui.developer", true) && autoload != null && autoload.trim().length() > 0) {
 				m_contextMaker = new ReloadingContextMaker(m_applicationClassName, m_config, autoload, autoloadWatchOnly);
-			else
+				if(null == m_isProductionMode)
```
fjalvingh: This is really odd logic. You are inside the filter's initialization so there should be no need to test for null of that field.

Luckily it needs to be replaced here anyway; you should set the new m_testMode field to true at this point- that is all.

to.etc.domui/src/main/java/to/etc/domui/server/ApplicationRequestHandler.java
```diff
@@ -50,7 +50,10 @@
 	private final DomApplication m_application;
 
 	@NonNull
-	private OopsFrameRenderer m_oopsRenderer = new OopsFrameRenderer();
+	private final OopsFrameRenderer m_oopsRenderer = new OopsFrameRenderer();
+
+	@NonNull
+	private final OopsFrameRenderer m_oopsPrdRenderer = new OopsFramePrdRenderer();
```
fjalvingh: I think this is not a very nice solution. Why not just pass the "testmode" as a parameter to the generation function in the OopsRenderer?

fjalvingh (changes requested):

to.etc.domui/src/main/java/to/etc/domui/server/OopsFrameRenderer.java
```diff
@@ -125,41 +133,46 @@ static private void dumpSingle(@NonNull StringBuilder sb, @NonNull Throwable x,
...
 
-	private static List<StackTraceElement> stripFrames(@NonNull List<StackTraceElement> list, int from) {
+	private List<StackTraceElement> stripFrames(@NonNull List<StackTraceElement> list, int from) {
 		return list.subList(0, from - 1);
 	}
 
-	private static void appendTraceLink(@NonNull StringBuilder sb, @NonNull StackTraceElement ste) {
-		sb.append("        <a class='exc-stk-l' href=\"#\" onclick=\"linkClicked('");
+	private void appendTraceLink(@NonNull StringBuilder sb, @NonNull StackTraceElement ste) {
+		if(m_stacktraceTag == "a") {
```
fjalvingh: That is an incorrect "if". The string "a" is a reference type, and m_stacktraceTag can easily point to another instance of the same value "a". In that case the if would fail. You need to use equals.
Using a tag name in this way also makes the code fragile because the decision what to render is for 99% here- but you have that tagname set outside. If you see how many ifs are inside here it is way better to just have a boolean m_testMode field and do the ifs on that. You can then make a local variable "tagName" in here if you really want (although for me doing without it would still be cleaner).
The same goes for the template name too, actually- way better to just select the correct template depending on m_testMode inside getExceptionTemplate().

fjalvingh (changes requested):
One small thing left, the rest is OK. Try to know when you can use == and when you must use equals.

fjalvingh (approved):

PR #306, opened 2021-12-14
Fixes missing dependency to okio

millevlada:


to.etc.domui.demo/pom.xml
```diff
@@ -137,6 +137,11 @@
             <!--<artifactId>jetty-servlets</artifactId>-->
         <!--</dependency>-->
 
+        <dependency>
+            <artifactId>okio</artifactId>
+            <groupId>com.squareup.okio</groupId>
+        </dependency>
```
fjalvingh, line 143: This should be a test dependency.

fjalvingh (approved):

PR #307, opened 2021-12-16
Delta minutes

buzakovica:


common/to.etc.alg/src/main/java/to/etc/util/DateUtil.java
```diff
@@ -590,6 +590,24 @@ public static int deltaInDays(Date start, Date end) {
...
 		return (int) ((double) (endL - startL) / (double) (1000 * 60 * 60 * 24));
 	}
+
+	/**
+	 * Get the difference in number of minutes between the specified dates.
+	 * @param start
+	 * @param end
+	 * @return
+	 */
+	public static int deltaInMinutes(Date start, Date end) {
```
millevlada, line 600: hm, why we needed this? Should be simple as:
TimeUnit.MILLISECONDS.toMinutes(start.getTime() - end.getTime());

Wherever that was needed, just add it inline as this. No need for method here. What you saw here was some older code, it is not needed now with these newer java abilities.

millevlada (changes requested):
this change is not needed, see remark

common/to.etc.alg/src/main/java/to/etc/util/DateUtil.java
buzakovica: You're right. Then I suggest deleting the PR?

millevlada:
decided to remove this branch

PR #352, opened 2022-03-30
added stripAccents method

buzakovica:


common/to.etc.alg/src/test/java/to/etc/alg/test/StringToolTest.java
```diff
@@ -0,0 +1,23 @@
+package to.etc.alg.test;
```
millevlada: should be 
to.etc.util

millevlada (approved):

PR #366, opened 2022-06-06
Added util to easily unzip zip with single file.

millevlada:


common/to.etc.alg/src/main/java/to/etc/util/FileTool.java
```diff
@@ -1354,6 +1354,44 @@ public static void unzip(File dest, InputStream is) throws Exception {
...
+			while(0 < (sz = zis.read(buf)))
+				os.write(buf, 0, sz);
+			os.close();
+			os = null;
+			zis.closeEntry();
+			if(null != zis.getNextEntry()) {
+				throw new IllegalArgumentException("Expected single file in zip but mutiple zip entries located!");
+			}
+		} finally {
+			if(zis != null)
```
fjalvingh, line 1382: Use try-with-resources wherever possible.

fjalvingh (approved):

PR #377, opened 2022-07-08
Added some fixes in sql date to util date conversion for in memory qu…

millevlada:
…eries.

### DIFF
--- a/to.etc.domui/src/main/java/to/etc/domui/util/db/CriteriaMatchingVisitor.java
+++ b/to.etc.domui/src/main/java/to/etc/domui/util/db/CriteriaMatchingVisitor.java
@@ -9,6 +9,7 @@ import to.etc.domui.component.meta.PropertyRelationType;
 import to.etc.domui.component.meta.impl.PathPropertyMetaModel;
 import to.etc.domui.util.DomUtil;
 import to.etc.domui.util.compare.StringLikeSearchMatchUtil;
+import to.etc.util.DateUtil;
 import to.etc.util.RuntimeConversions;
 import to.etc.webapp.qsql.QQuerySyntaxException;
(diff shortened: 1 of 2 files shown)


to.etc.domui/src/main/java/to/etc/domui/util/db/CriteriaMatchingVisitor.java
```diff
@@ -95,6 +96,14 @@ public void visitPropertyComparison(@NonNull QPropertyComparison n) throws Excep
 		Class< ? > litc = DomUtil.getUnproxiedClass(lit.getClass());
 		Class< ? > valc = DomUtil.getUnproxiedClass(val.getClass());	// Types differ?
 		if(litc != valc) {
+			if(valc.isAssignableFrom(java.sql.Date.class) && litc.isAssignableFrom(java.util.Date.class)) {
```
millevlada, line 99: Is this ok, or we like to have some more generic approach?

fjalvingh, line 99: The first thing should be an instanceof, not isAssignableFrom

fjalvingh: If you look at the code the proper thing to do would be to use the getPromoted method and all. Let me see how bad that would get.

millevlada: is this now fixed with alternative solution? Then we should close this PR as obsolete.

fjalvingh:
Yes, I have used the existing mechanism to promote both types to the same date type (just added the conversions, and added the promotions to the order list).

PR #398, opened 2022-09-23
Added option for custom value renderer in RadioGroup.

millevlada:
Added EnumValueListInput component. It enables defining list of values that have given enum value as values.


to.etc.domui/src/main/java/to/etc/domui/dom/html/RadioGroup.java
```diff
@@ -158,9 +163,22 @@ public RadioButton<T> addButton(String text, T value, @Nullable String title) {
 		add(d);
 		RadioButton<T> rb = new RadioButton<>(value);
 		d.add(rb);
-		Label label = new Label(rb, text);
-		d.add(label);
-		label.setTitle(title);
+		IRenderInto<ValueLabelPair<T>> valueRenderer = getValueRenderer();
+		if(null != valueRenderer) {
```
fjalvingh, line 167: Invert that condition?

fjalvingh (approved):

PR #413, opened 2022-11-14
Added option to add a check if tab can be selected.

millevlada:


to.etc.domui/src/main/java/to/etc/domui/component/layout/TabPanelBase.java
```diff
@@ -60,6 +60,13 @@ public interface ITabSelected {
 		void onTabSelected(TabPanelBase tabPanel, int oldTabIndex, int newTabIndex) throws Exception;
 	}
 
+	/**
+	 * Represents check if tab can be selected.
+	 */
+	public interface ITabSelectable {
+		boolean isTabSelectable(TabPanelBase tabPanel, int oldTabIndex, int newTabIndex) throws Exception;
```
fjalvingh: Use TabInstance instead of int to indicate the from/to change, do not use indices. I also think that it is better not to add the tabpanel itself; that type of parameter is almost never used and is very easy to provide in the listener itself.

to.etc.domui/src/main/java/to/etc/domui/component/layout/TabPanelBase.java
```diff
@@ -428,15 +450,24 @@ public int getTabCount() {
...
+	public void setOnTabSelected(@Nullable ITabSelected onTabSelected) {
 		m_onTabSelected = onTabSelected;
 	}
 
+	@Nullable
 	public ITabSelected getOnTabSelected() {
 		return m_onTabSelected;
 	}
 
+	public void setOnTabSelectable(@Nullable ITabSelectable onTabSelectable) {
```
fjalvingh, line 469: Add a Javadoc here, and also tell there that the handler MUST give some kind of indication of WHY switching tabs is not possible, otherwise you will just have a thing that does not react when clicked which is a Bad Thing(tm).

fjalvingh (changes requested):

fjalvingh (approved):

PR #428, opened 2022-12-26
Added whitelisting override for sql name checks.

millevlada:


fjalvingh:
We really do not need nor really want that. I changed the code that called that.

PR #432, opened 2023-01-06
Enables sorting/non sorting of enum members when creating the input a…

millevlada:
…s radio groups. Backward compatible is sorted.

### DIFF
--- a/to.etc.domui/src/main/java/to/etc/domui/dom/html/RadioGroup.java
+++ b/to.etc.domui/src/main/java/to/etc/domui/dom/html/RadioGroup.java
@@ -215,7 +215,11 @@ public class RadioGroup<T> extends Div implements IHasChangeListener, IControl<T
 		return createEnumRadioGroup(clz, true, valueRenderer, exceptions);
 	}
 
-	static public <T extends Enum<T>> RadioGroup<T> createEnumRadioGroup(Class<T> clz, boolean sorted, @Nullable IRenderInto<ValueLabelPair<T>> valueRenderer, T... exceptions) {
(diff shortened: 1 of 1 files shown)


to.etc.domui/src/main/java/to/etc/domui/dom/html/RadioGroup.java
```diff
@@ -212,6 +212,10 @@ static public <T extends Enum<T>> RadioGroup<T> createEnumRadioGroup(Class<T> cl
 	}
 
 	static public <T extends Enum<T>> RadioGroup<T> createEnumRadioGroup(Class<T> clz, @Nullable IRenderInto<ValueLabelPair<T>> valueRenderer, T... exceptions) {
+		return createEnumRadioGroup(clz, true, valueRenderer, exceptions);
+	}
+
+	static public <T extends Enum<T>> RadioGroup<T> createEnumRadioGroup(Class<T> clz, boolean sorted, @Nullable IRenderInto<ValueLabelPair<T>> valueRenderer, T... exceptions) {
```
fjalvingh: I do not like boolean parameters in public interface methods. Can we:
- make this method private
- add another public method createEnumRadioGroupUnsorted() instead?

fjalvingh (changes requested):

to.etc.domui/src/main/java/to/etc/domui/dom/html/RadioGroup.java
millevlada: ok, fixed

fjalvingh (approved):

PR #442, opened 2023-01-16
Added ActionButton.

millevlada:
Added disabled reason for PopupMenu2 options, so it can better reuse IUIAction. Also added options to use text and hint as plain text. Fixed z-index for popup menu given inside floating panels.

### DIFF
--- a/to.etc.domui/src/main/java/to/etc/domui/component/buttons/ActionButton.java
+++ b/to.etc.domui/src/main/java/to/etc/domui/component/buttons/ActionButton.java
@@ -7,6 +7,7 @@ import to.etc.domui.component.misc.Icon;
 import to.etc.domui.component2.popupmenus.PopupMenu2;
 import to.etc.domui.dom.html.HR;
 import to.etc.domui.dom.html.NodeBase;
+import to.etc.util.Pair;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


to.etc.domui/src/main/java/to/etc/domui/component/buttons/ActionButton.java
```diff
@@ -0,0 +1,68 @@
...
+	@Override
+	public void createContent() throws Exception {
+		super.createContent();
+		if(m_actions.isEmpty()) {
+			return;
+		}
+		HR hr = new HR();
+		hr.addCssClass("act-rule");
+		add(hr);
+		NodeBase actionButton = Icon.faEllipsisV.createNode();
```
fjalvingh: If the button is disabled the menu needs to be disabled too.

to.etc.domui/src/main/java/to/etc/domui/component/buttons/ActionButton.java
```diff
@@ -0,0 +1,68 @@
...
+	private final List<IUIAction<T>> m_actions = new ArrayList<>();
+
+	private final T m_instance;
+
+	public ActionButton(T instance, IUIAction<T> action) throws Exception {
+		super(action);
+		m_instance = instance;
+	}
+
+	public ActionButton<T> addAction(IUIAction<T> action) {
```
fjalvingh: Why would all instances need to be a "T"? I think both the primary action and the additional actions could have different instances?

to.etc.domui/src/main/java/to/etc/domui/component/buttons/ActionButton.java
fjalvingh: On a button of this type the icon is not a ... but a v (down triangle)

fjalvingh (changes requested):

to.etc.domui/src/main/java/to/etc/domui/component/buttons/ActionButton.java
millevlada: ok, fixed

to.etc.domui/src/main/java/to/etc/domui/component/buttons/ActionButton.java
millevlada: fixed

fjalvingh (approved):

PR #487, opened 2023-06-02
Added meta utils to mark secret props that can be stored as encrypted…

millevlada:
… where needed in generic prop


fjalvingh:
Should no longer be needed, we have an annotation to handle this at the application level.

PR #490, opened 2023-06-13
added test helper methods

sionans:


integrations/to.etc.domui.selenium/src/main/java/to/etc/domui/webdriver/poproxies/CpComboLookup2.java
```diff
@@ -10,4 +11,10 @@ public class CpComboLookup2 extends AbstractCpComboComponent implements ICpContr
 	public CpComboLookup2(WebDriverConnector connector, Supplier<String> selectorProvider) {
 		super(connector, selectorProvider);
 	}
+
+	public Select getOptions() {
```
fjalvingh: As this returns the "select" it should be called getSelect(). Also because the things "below" a <select> tag, which contains the options, is called <option>.

fjalvingh (changes requested):

fjalvingh (approved):

PR #508, opened 2023-10-20
extending component EnumValueListInputBase

milosilic:
Allows you to pass additional checks whether control should be readonly or whether some enum should be ignored


fjalvingh (changes requested):
I am not going to accept this.
This adds methods that "pull" information from the main program without any need. We only use "pulling" methods if there is no alternative. The normal  way of configuring controls is by pushing info into them, i.e. calling setReadOnly instead of calling a method that then decides whether the thing is readonly.

The same goes for the enum filter: if you do not want all enum members then provide a list of members that should be shown (for instance in the constructor). Do not add a method that needs to filter; what needs to be shown is the task of the caller.

PR #509, opened 2023-11-07
workaround for ssl issues with nltaxonomie.nl

millevlada:


common/to.etc.alg/src/main/java/to/etc/net/SslParameters.java
```diff
@@ -0,0 +1,63 @@
+package to.etc.net;
```
millevlada: I'd like to move this class here from shared command line (in Skarp) code.

common/to.etc.alg/src/main/java/to/etc/net/HttpSsl.java
```diff
@@ -0,0 +1,141 @@
...
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.Executor;
+import java.util.concurrent.ExecutorService;
+
+public class HttpSsl {
```
millevlada: some parts of this code exist also as JdkHttpClient (in Skarp code). But if we do not use GenericHttpRequest abstraction, then we do not need these other parts... So, I placed here parts that could be used for simplified use of ssl client.

integrations/to.etc.domui.hibutil/src/main/java/to/etc/domui/hibernate/config/HibernateConfigurator.java
```diff
@@ -446,6 +447,8 @@ public synchronized static void initialize(final DataSource ds) throws Exception
 		}
 
 		m_contextSource = new HibernateLongSessionContextFactory(m_listeners, hsm, m_handlers, HibernateConfigurator::onContextCreated);
+		ExceptionDialog.register(HibernateMessageDecoder::translateHibernateException);
```
millevlada: No idea how this ended up here, I guess that is something I picked up with some merges? It does look like some valid change, so I'll assume it is ok?

millevlada: yes, it went into domui meanwhile, so it is ok

fjalvingh (approved):

PR #510, opened 2023-11-10
delete Temp files on clear button pressed

milosilic:
When a file is uploaded using FileUpload2 or FileUploadMultiple and then cleared from the element using clear button, it will remain hanging in temp folder of server. 
This PR should fix that

### DIFF
--- a/to.etc.domui/src/main/java/to/etc/domui/component/upload/FileUpload2.java
+++ b/to.etc.domui/src/main/java/to/etc/domui/component/upload/FileUpload2.java
@@ -49,6 +49,7 @@ import to.etc.domui.util.DomUtil;
 import to.etc.domui.util.Msgs;
 import to.etc.domui.util.upload.FileUploadException;
 import to.etc.domui.util.upload.UploadItem;
+import to.etc.util.FileTool;
 
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


to.etc.domui/src/main/java/to/etc/domui/component/upload/FileUpload2.java
```diff
@@ -313,6 +314,7 @@ private void clearClicked() throws Exception {
 	public void clear() throws Exception {
 		if(m_value == null)
 			return;
+		FileTool.closeAll(m_value.getFile());
```
millevlada, line 317: since it might be unclear why is that needed, we could add an comment next to it:
//removes created temp file

to.etc.domui/src/main/java/to/etc/domui/component/upload/FileUploadMultiple.java
```diff
@@ -191,6 +192,7 @@ private void renderValue(Div valueD, UploadItem uploadItem) {
 		d.add(uploadItem.getRemoteFileName());
 		btn.setClicked(a -> {
 			m_value.remove(uploadItem);
+			FileTool.closeAll(uploadItem);
```
millevlada, line 195: same here, comment would help to understand why this code is useful

to.etc.domui/src/main/java/to/etc/domui/component/upload/FileUploadMultiple.java
```diff
@@ -245,10 +247,10 @@ private void validate() {
 		}
 	}
 
-	@Override public void setValue(@Nullable List<UploadItem> value) {
+	@Override
```
millevlada, line 250: this was not necessary change, we try not to change unchanged code

fjalvingh (approved):

PR #511, opened 2023-11-13
register uploaded items for delete

milosilic:
If m_isRegisterAutoCleanup is set, all files that are uploaded in temp folder of the server using FileUploadMultiple component will be deleted at end of conversation time


to.etc.domui/src/main/java/to/etc/domui/component/upload/FileUploadMultiple.java
```diff
@@ -345,6 +347,9 @@ public boolean handleUploadRequest(@NonNull RequestContextImpl param, @NonNull C
 			if(uiar != null) {
 				for(UploadItem ui : uiar) {
 					m_value.add(ui);
+					if(m_isRegisterAutoCleanup){
+						conversation.registerTempFile(ui.getFile());
```
fjalvingh, line 351: getFile() can return null in which case it should not be passed to registerTempFile().

fjalvingh:
I did a similar fix but everywhere where it was needed, so this PR is no longer needed.

PR #513, opened 2023-11-28
new helper method

milosilic:
Added new method 'isUnique' to QQueryUtils to assist in determining the uniqueness of a propertyin database

### DIFF
--- a/to.etc.webapp.core/src/main/java/to/etc/webapp/query/QQueryUtils.java
+++ b/to.etc.webapp.core/src/main/java/to/etc/webapp/query/QQueryUtils.java
@@ -48,12 +48,13 @@ final public class QQueryUtils {
 		return pf.createInstance(resl);
 	}
 
-	public static <T extends IIdentifyable<?>> boolean isUnique(@NonNull QDataContext dc, @NonNull QCriteria<T> qc, T object)
-		throws Exception {
-        if(null != object.getId()){
-			qc.ne("id", object.getId());
+
+	public static <T extends IIdentifyable<?>, V> boolean isUnique(@NonNull QDataContext dc, Class<T> clz, @NonNull QField<T, V> property, V value, @NonNull QField<T, V> propertyId, V id ) throws Exception {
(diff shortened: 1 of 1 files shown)


to.etc.webapp.core/src/main/java/to/etc/webapp/query/QQueryUtils.java
```diff
@@ -48,6 +48,15 @@ static public <R> R mapSelectionOneQuery(@NonNull QDataContext dc, @NonNull Clas
 		return pf.createInstance(resl);
 	}
 
+
+	public static <T extends IIdentifyable<?>, V> boolean isUnique(@NonNull QDataContext dc, Class<T> clz, @NonNull QField<T, V> property, V value, @NonNull QField<T, V> propertyId, V id ) throws Exception {
```
millevlada, line 52: No need to pass in id as separate value.
You should be able to get it with:
V id = (V) value.getId();

To avoid cast, we can also use:
PropertyMetaModel<T> pmm = MetaManager.getPropertyMeta(clz, propertyId);
V id = pmm.getValue(value);

millevlada (changes requested):
see remarks

to.etc.webapp.core/src/main/java/to/etc/webapp/query/QQueryUtils.java
milosilic: PropertyMetaModel and MetaManager are in to.etc.domui module...

fjalvingh (approved):

to.etc.webapp.core/src/main/java/to/etc/webapp/query/QQueryUtils.java
millevlada: ok, make new class in to.etc.domui.util.db, QChecksUtil, and move the code there.
Also, fix the generic types in this, use P for id type, and then we should be able to work with less parameters over all, since both id and property value to compere can be taken from passed in value object.

PR #525, opened 2024-02-01
Add isAcceptable method for filtering and better item selection logic

milosilic:
Enhancement on  DataTable and InstanceSelectionModel classes to improve item filtering and selection logic.

In InstanceSelectionModel, a new method isAcceptable(T instance) has been added. This method allows for a customizable check to determine if an row meets certain criteria before being accepted. 

In DataTable, the selection/unselect logic has been improved with isAcceptable method.

### DIFF
--- a/to.etc.domui/src/main/java/to/etc/domui/component/tbl/DataTable.java
+++ b/to.etc.domui/src/main/java/to/etc/domui/component/tbl/DataTable.java
@@ -129,8 +129,14 @@ final public class DataTable<T> extends PageableTabularComponentBase<T> implemen
 			//in case of multiselection, if not all visible items in model are selected, add to selection, otherwise do clean of selection
 			ITableModel<T> model = getModel();
 			boolean performSelect = sm.getSelectionCount() == 0;
-			if(! performSelect) {
-				performSelect = model.getItems(0, model.getRows()).stream().anyMatch(it -> !sm.isSelected(it));
+			if(!performSelect) {
+				performSelect = model.getItems(0, model.getRows()).stream().anyMatch(it -> {
+					boolean isAcceptable = true;
+					if(sm instanceof InstanceSelectionModel) {
+						isAcceptable = ((InstanceSelectionModel<T>) sm).isAcceptable(it);
+					}
+					return isAcceptable && !sm.isSelected(it);
+				});
 			}
 			if(performSelect) {
 				sm.selectAll(model);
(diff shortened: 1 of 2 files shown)


to.etc.domui/src/main/java/to/etc/domui/component/tbl/DataTable.java
```diff
@@ -129,8 +129,14 @@ final public class DataTable<T> extends PageableTabularComponentBase<T> implemen
 			//in case of multiselection, if not all visible items in model are selected, add to selection, otherwise do clean of selection
 			ITableModel<T> model = getModel();
 			boolean performSelect = sm.getSelectionCount() == 0;
-			if(! performSelect) {
-				performSelect = model.getItems(0, model.getRows()).stream().anyMatch(it -> !sm.isSelected(it));
+			if(!performSelect) {
+				performSelect = model.getItems(0, model.getRows()).stream().anyMatch(it -> {
+					boolean isAcceptable = true;
+					if(sm instanceof InstanceSelectionModel) {
```
fjalvingh, line 135: No, this does not work. You have this problem in ALL selection models, so you would need to put that isAcceptable() method in the interface. In general casting to a specific interface implementation in a generic component is very bad form.

There is also a technical issue here that you could not know about.
This code must work the same for all selection models. But some Data Models are actually very lazy: they do not contain instances but only keys, and only when we really need an instance (because we want to show it in the list) will that instance actually be created.

The code here causes a large problem for these kinds of models because it forces the model to instantiate ALL instances- and that can be hideously expensive and even lead to OOMs.

Instead of this loop here which instantiates all items you need to have something like isAnySelected(model) inside the SelectionModel interface, so that the implementation of the model can decide what is the best way to handle this.

fjalvingh: Actually, I think we have a slightly different issue too here.
We have a selection model that can hold items that are "outside" (not present) of the currently visible Model of the table. The first rule we use here (I think correctly) is:
* If nothing of the current Model is selected we want to SELECT items
* If one or more items are selected we want to CLEAR items

The "SELECT ITEMS" is done correctly: we call selectAll() with the model, and this will select all items inside the model.
But the CLEAR ITEMS thing seems wrong: it clears ALL items in the model, not only the items that are in the currently visible model. I think that that is incorrect. I think it should clear only the items that came from the current model....

I will ask Vladimir's and Jovana's opinion so wait with fixing until we have a clear idea.

fjalvingh (changes requested):

fjalvingh:
I rewrote the implementation for both keyed and instance models.

PR #541, opened 2024-03-13
localization of enums in PropertyExportColumn

milosilic:
Localization on PropertyExportColumn exporter is being ignored on enums. 
This is a fix for that

### DIFF
--- a/to.etc.domui/src/main/java/to/etc/domui/util/exporters/PropertyExportColumn.java
+++ b/to.etc.domui/src/main/java/to/etc/domui/util/exporters/PropertyExportColumn.java
@@ -3,8 +3,10 @@ package to.etc.domui.util.exporters;
 import org.eclipse.jdt.annotation.NonNull;
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
+import to.etc.domui.component.meta.ClassMetaModel;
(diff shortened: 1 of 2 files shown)


to.etc.domui/src/main/java/to/etc/domui/util/exporters/PropertyExportColumn.java
```diff
@@ -47,4 +49,17 @@ public PropertyExportColumn(PropertyMetaModel<T> pmm) {
 	@Nullable @Override public T getValue(@NonNull Object in) throws Exception {
 		return m_pmm.getValue(in);
 	}
+
+	@Nullable
+	@Override
+	public Object convertValue(@Nullable Object value) throws Exception {
+		if(value instanceof Enum) {
+			if(value == null)
```
fjalvingh, line 57: If you have instanceof the result after it cannot be null: if value was null then that instanceof would not be true because null is not an instance of any class according to the definition of instanceof.

fjalvingh (approved):

to.etc.domui/src/main/java/to/etc/domui/util/exporters/PropertyExportColumn.java
fjalvingh: You should also see that in the IDE; it should have marked that if as something problematic.

PR #544, opened 2024-07-19
Fix Excel number formatting to prevent rounding in exported files

milosilic:
Fix Excel number formatting to prevent rounding in exported files

- Updated number cell style to use "#,##0.###############" format for correct display of decimal values.
- Ensured that numbers less than 1 are displayed with a leading zero.

### DIFF
--- a/to.etc.domui/src/main/java/to/etc/domui/util/exporters/ExcelExportWriter.java
+++ b/to.etc.domui/src/main/java/to/etc/domui/util/exporters/ExcelExportWriter.java
@@ -63,6 +63,9 @@ public class ExcelExportWriter<T> implements IExportWriter<T> {
 	private CellStyle m_numberStyle;
 
 	@Nullable
+	private CellStyle m_decimalStyle;
+
+	@Nullable
 	private CellStyle m_headerStyle;
 
 	@Nullable
@@ -112,6 +115,10 @@ public class ExcelExportWriter<T> implements IExportWriter<T> {
 		CellStyle nums = m_numberStyle = wb.createCellStyle();
 		nums.setDataFormat(wb.createDataFormat().getFormat("#0"));
 		nums.setFont(defaultFont);
+
+		CellStyle dums = m_decimalStyle = wb.createCellStyle();
+		dums.setDataFormat(wb.createDataFormat().getFormat("#,##0.####"));
+		dums.setFont(defaultFont);
 
 		Font headerFont = wb.createFont();
 		headerFont.setFontHeightInPoints((short) 10);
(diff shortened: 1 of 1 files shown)


to.etc.domui/src/main/java/to/etc/domui/util/exporters/ExcelExportWriter.java
```diff
@@ -367,18 +385,18 @@ public void renderCell(ExcelExportWriter<?> exporter, Cell cell, int cellIndex,
 		register(byte.class, NUMBER_CONVERTER);
 		register(short.class, NUMBER_CONVERTER);
 		register(int.class, NUMBER_CONVERTER);
-		register(long.class, NUMBER_CONVERTER);
-		register(float.class, NUMBER_CONVERTER);
+		register(long.class, DECIMAL_CONVERTER);
```
fjalvingh: long cannot have a fraction, it should remain NUMBER_CONVERTER.

to.etc.domui/src/main/java/to/etc/domui/util/exporters/ExcelExportWriter.java
```diff
@@ -367,18 +385,18 @@ public void renderCell(ExcelExportWriter<?> exporter, Cell cell, int cellIndex,
...
+		register(float.class, DECIMAL_CONVERTER);
 		register(double.class, NUMBER_CONVERTER);
```
fjalvingh: Double, however, has a fraction, so it needs DECIMAL_CONVERTER.

to.etc.domui/src/main/java/to/etc/domui/util/exporters/ExcelExportWriter.java
```diff
@@ -321,6 +328,17 @@ static public <V> IExportCellRenderer<ExcelExportWriter<?>, Cell, V> findRendere
 		@Override
 		public void renderCell(ExcelExportWriter<?> exporter, Cell cell, int cellIndex, @Nullable Number value) throws Exception {
 			cell.setCellStyle(exporter.m_numberStyle);
+			if(null != value) {
+				cell.setCellValue(value.longValue());
```
fjalvingh: This is wrong; if the value is BigInteger it could easily not fit. You can cast the rest to long, but BigInteger needs to be set using doubleValue. Use instanceof to test.

fjalvingh (changes requested):

to.etc.domui/src/main/java/to/etc/domui/util/exporters/ExcelExportWriter.java
milosilic: ofc

fjalvingh (approved):

PR #557, opened 2024-08-29
method to set label before checkbox in CheckboxSetInputBase class

milosilic:
﻿Ability to set the labels for each checkbox  on the left side of the toggle.
﻿The spacing between two filters and their belonging labels are made bigger so it is more clear which label refers to which toggle.

### DIFF
--- a/to.etc.domui/src/main/java/to/etc/domui/component/input/CheckboxSetInputBase.java
+++ b/to.etc.domui/src/main/java/to/etc/domui/component/input/CheckboxSetInputBase.java
@@ -113,6 +113,7 @@ abstract public class CheckboxSetInputBase<V, T> extends AbstractDivControl<Set<
 		Label span = new Label();
 		span.setForTarget((NodeBase) cb);
 		if(m_labelBeforeCheckbox) {
+			span.addCssClass("label-first");
 			pair.add(span);
 			pair.add((NodeBase) cb);
 		} else {
--- a/to.etc.domui/src/main/java/to/etc/domui/component/input/CheckboxSetInputBase.java
+++ b/to.etc.domui/src/main/java/to/etc/domui/component/input/CheckboxSetInputBase.java
@@ -34,6 +34,8 @@ abstract public class CheckboxSetInputBase<V, T> extends AbstractDivControl<Set<
 
 	private boolean m_asButtons;
 
+	private boolean m_labelBeforeCheckbox;
+
 	@NonNull
 	abstract protected V listToValue(@NonNull T in) throws Exception;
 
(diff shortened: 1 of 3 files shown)


to.etc.domui/src/main/resources/resources/themes/scss/winter/_checkboxset.scss
```diff
@@ -7,6 +7,11 @@
 
 		.ui-cbis-p {
 			padding-right: 10px;
+			display: flex;
+			align-items: center;
+		}
+		.ui-cbis-p label {
```
millevlada: We want extra margin only when label is before checkbox, otherwise it would consume some unwanted space when it is used as after checkbox.
Add separate part of style (label-first) to mark that:
.ui-cbis-p {
  padding-right: 10px;
  &.label-first {
    display: flex;
    align-items: center;
    label {
      margin-right: 4px;
    }
  }
}

to.etc.domui/src/main/java/to/etc/domui/component/input/CheckboxSetInputBase.java
```diff
@@ -105,13 +107,18 @@ private void renderCheckbox(Div setContainer, @NonNull T lv) throws Exception {
...
 
-		pair.add((NodeBase) cb);
 		IRenderInto<T> cr = m_actualContentRenderer;
 		if(cr == null)
 			cr = m_actualContentRenderer = calculateContentRenderer(lv);
 		Label span = new Label();
 		span.setForTarget((NodeBase) cb);
-		pair.add(span);
+		if(m_labelBeforeCheckbox) {
+			pair.add(span);
```
millevlada, line 117: add:
span.addCss("label-first");

millevlada (changes requested):
see remarks

millevlada (approved):

to.etc.domui/src/main/java/to/etc/domui/component/input/CheckboxSetInputBase.java
```diff
@@ -105,13 +107,19 @@ private void renderCheckbox(Div setContainer, @NonNull T lv) throws Exception {
...
 
-		pair.add((NodeBase) cb);
 		IRenderInto<T> cr = m_actualContentRenderer;
 		if(cr == null)
 			cr = m_actualContentRenderer = calculateContentRenderer(lv);
 		Label span = new Label();
 		span.setForTarget((NodeBase) cb);
-		pair.add(span);
+		if(m_labelBeforeCheckbox) {
+			span.addCssClass("label-first");
```
fjalvingh, line 116: All css classes in DomUI use the ui-prefix combined with the component short code.

fjalvingh (approved):

PR #585, opened 2024-12-04
title attribute is not supported on TextNode element

milosilic:
Title attrubutes should now be supported on TextNode element


fjalvingh (approved):

fjalvingh:
While true there is a sh*tload of attributes that are present on textnode that cannot be set. It is actually a flaw in the design: TextNode should not extend NodeBase but some other, intermediary class. Adding only a check on title does not seem worthwhile.

PR #610, opened 2025-08-14
transient values skipped when copying values from object

milosilic:


fjalvingh (changes requested):
No, that is not reasonable. There is no specific reason why copying transients is bad, in fact it usually is a good idea to copy them.

